PR #272, opened 2015-07-05
Add M3/M5 Buttontoactivate/deactivate Spindle (For Laser orother)

Crazydoub:
It would begreat to Add M3 & M5 Buttontoactivate/deactivate Spindle (For Laser or other) like found on GRBL controller.
Sometimeswhen my g-code is not great, laser still enabled after work so i could stop it :D


winder:
I'm not sure what this pull request is, could you re-create it as an enhancement request?

Thanks

PR #277, opened 2015-07-12
declare compiler source and target in pom.xml to 1.7

nitram509:
Hi,

as it is required since UGS 1.7, the Java source and target in the pom.xml should also be 1.7.
I also added IntelliJ IDEA project files to the ignore list.

Regards
Martin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ tar.gz
 /UGSPlatform/CentralLookup/nbproject/private/
 /UGSPlatform/Options/nbproject/private/
 /UGSPlatform/VisualizerModule/build/
+.idea
+*.iml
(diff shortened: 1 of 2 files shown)


nitram509:
@winder 

I saw an Ant build.xml and a Maven pom.xml.
Why there are two build configurations?
Do you have a wish, which one to choose for the future?

Regards
Martin

PS: Personally, I prefer Maven ;)

winder:
The pom.xml file isn't officially supported for a few reasons:
- I couldn't get it to work with one-jar, so I can't make release builds with it.
- I need it to be compatible with Netbeans so that I can use the Swing GUI Builder.
- I couldn't get the new UGS Platform project to build properly with maven.

PR #281, opened 2015-07-18
Fixed Maven pom.xml - now full compatible

nitram509:
Hi,

I've fixed the pom.xml.
Now Maven is able to compile, test, package and build a release file (ZIP).

The shortest path to create a UniversalGcodeSender.zip is:
mvn package assembly:assembly

Just run tests:
mvn test

Just build a self-running uber-jar:
mvn package

Because of maven support, some folders may be deleted now:
lib, test_lib, release, nbproject, BundlingTools, UGSPlatform, build.xml, manifest.mf, one-jar-ant-task.xml, onejar.mf

What do you think about fully switching to Maven?

Regards
Martin


winder:
I would like to switch over to maven for everything. There are still some things missing though - its a lot closer than my previous attempts though.

Could you explain the uber-jar / shade plugin? Is this replacing one-jar? I switched away from RXTX to JSSC, so I think all of the libraries with native dependencies come pre-packaged now which could eliminate the one-jar dependency.
I'm currently using a customized version of one-jar to extract native binaries in the lib/natives directories (the customization added support for raspberry pi). The uber-jar worked on a windows VM but not on my Linux box - I tried a 32-bit and 64-bit JRE and both reported incompatible shared library errors:
```
### With a 32-bit JRE
Exception in thread "AWT-EventQueue-0" java.lang.UnsatisfiedLinkError: /tmp/jogamp_0000/file_cache/jln90510860563301329/jln6016019128624081994/libgluegen-rt.so: /tmp/jogamp_0000/file_cache/jln90510860563301329/jln6016019128624081994/libgluegen-rt.so: cannot open shared object file: No such file or directory (Possible cause: can't load ARM-bit .so on a IA 32-bit platform)
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1937)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1822)

### With a 64-bit JRE
Exception in thread "AWT-EventQueue-0" java.lang.UnsatisfiedLinkError: /tmp/jogamp_0000/file_cache/jln6462992554056791763/jln7291808247666853734/libgluegen-rt.so: /tmp/jogamp_0000/file_cache/jln6462992554056791763/jln7291808247666853734/libgluegen-rt.so: wrong ELF class: ELFCLASS32 (Possible cause: architecture word width mismatch)
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1937)
```


Another challenge is that I started a parallel project that lives in the "UGSPlatform" directory which is a whole separate challenge. It is a "Netbeans Platform" front-end and is also using ANT because I couldn't get Maven to work properly for it either. The Maven support for Netbeans Platform applications seems to be limited right now.

In ant you can see the UGS Platform build with "ant run-platform" from the root directory (which kicks off the "run" target in the UGSPlatform directory.

nitram509:
Thank you for pointing the issue with the native files out.

> Could you explain the uber-jar / shade plugin? Is this replacing one-jar?
Regarding the onejar-maven-plugin I had a personal concern: last release was 2011.
Additionally on [JogAmp Wiki](https://jogamp.org/wiki/index.php/JogAmp_JAR_File_Handling)
was mentioned that one-jar lacks URL handler for jar-in-jar resources.
That read, I opted for maven-shade-plugin, because I have good experience with it.
Besides that, the shade-plugin is way smarter than the assembly plugin.
Cite from assembly-plugin site:
   "If your project wants to package your artifact in an uber-jar, the assembly plugin provides only basic support. For more control, use the Maven Shade Plugin."

To create a fat-jar/uber-jar (to me these are synonyms) the configuration of shade-plug in pom.xml is sufficient (see line 322 in pom.xml).
But as there is an additional file UniversalGcodeSender.zip required, so the assembly-plugin jumps in. This one is only responsible for creating the ZIP file.

> I switched away from RXTX to JSSC.
I also removed dependency RXTX in pom.xml

>  I tried a 32-bit and 64-bit JRE and both reported incompatible shared library errors
I oversaw, that the .so or .dll file names are the same on each architecture.
So I fixed it, by using the "natives" folder mechanism, as described at [JogAmp Wiki](https://jogamp.org/wiki/index.php/JogAmp_JAR_File_Handling).
I've tested on Windows, Linux and Mac OSX - everything worked fine.

nitram509:
Regarding your parallel project "UGSPlatform", I guess you plan to switch to this with version 2.0 and get rid of the classical UI, right?

winder:
For the UGS Platform yes and no, It uses the same UGS jar as a library. So
the classic build won't go away any time soon.

nitram509:
Maintaining two UIs means more effort. May I ask what are your reasons for?

winder:
There is an API for front ends to use, so I shouldn't need to maintain the
old UI. That's the plan anyway.

winder:
I renamed the build.xml file and nbproject directory and my IDE opens the project just fine. Great! I still need to figure out the UGSPlatform build though. Thanks for all your effort on this.

nitram509:
You're welcome.

Issue #168, opened 2014-06-09
Universal-G-Code-Sender becomes unresponsive after ~250,000 lines


alsliahona:
When using the Universal GCode Sender on a well powered Windows 8.1 PC everything runs fine until about 250,000 lines of gcode have been sent.  After this the program becomes somewhat unresponsive (clicks take many minutes to be processed) and the feed rate to GRBL is so slow that the end product ends up being severely damaged (especially when using a laser, because the movement is so slow as GRBL waits for a new command that it burns the current location deeply.)

I've tested several times and see this happen consistently after about 253,000 lines.  PicLaser Lite will easily generate a file with twice that number of lines for a small image.

Running the same GCode through Grbl Controller 3.0 (on Windows) worked fine.

It looks like it is probably a resource leak -- which because of Java (lame) garbage collection may be outside of the developers direct control.

As a side note, Universal GCode Sender is unusable on a Raspberry Pi because of the intense resource usage -- it will immediately slow down the machine.

winder:
To be addressed in #302

PR #299, opened 2015-10-07
performance enhancement - optimize file preprocessing

philreindl:
When opening large files, the preprocessing takes a very long time.  By compiling the regular expressions once rather than using the string.replaceAll methods, we can save significant time.

On my laptop, I opened a gcode file with 7233438 lines (laser image raster)
with my changes - 16155ms
master - 56178ms 

I believe this addresses at least part of Issue #168.


winder:
Thanks!

PR #311, opened 2015-11-11
Restrict keyboard-based jogging to sane moves

clockfort:
Maybe this should be hidden behind another setting?

Almost everyone who does the currently easy-to-get-into-with-the-keyboard-controls 10 or 100 inch jumps is guaranteed to just have a crash...
I haven't had one yet with my jog controller, but it's a matter of time with moves that big allowed...

(Related, thanks for an awesome project, I have a jog controller set up through UGS and it's pretty awesome https://www.youtube.com/watch?v=SO-YCPmuBhc )


faultylee:
+1, I get caught in this from time to time when the usb become unstable with grbl, though my problem was more of jogging Z in inches

clockfort:
There's probably some better more generic way to doing this feature. Maybe "These are my machine's work area dimensions, never let me jog more than 25% of that axis' full distance at a time, because I admit I'm probably being an idiot"

winder:
Sorry to close this PR after all this time. Limiting the jogging like this seemed kind of arbitrary so I wasn't sure about including it - especially since `MainWindow.java` is filled with so many arbitrary things already. On the platform build there is a `JogService.java` class to hold all these sorts of things, I'd be much more comfortable adding a change like this there.

PR #313, opened 2015-11-18
Removed the File Tab. New Panel with the File Controls

rugbymauri:
Removed the File Tab. New Panel with the File Controls
Moved the Job Infomation in the Machine Status Panel

universal gcode sender version 2 0 nightly - _024


winder:
@rugbymauri I don't want to change the classic GUI this much. It's one of those things that everyone is going to want a little different.

This sort of thing will be good for the platform project though, I think I even did something similar there. If you haven't checked it out recently, please take another look. I'm curious about any suggestions you have there.

Issue #302, opened 2015-10-14
Huge file support - UGS to use constant memory!


winder:
Many people have noticed that UGS doesn't support very large files. This is due to UGS storing several copies of the entire file in memory for the duration of an operation. Specifically there are 3 places that are storing the entire file in memory (with additional storage for metadata!):

1) Each line is stored in a GcodeCommand object, which has some additional metadata.
2) If the visualizer is being used, each vertex is stored in memory - for arcs this vertex data can be even larger than a single command..
3) The GUI even keeps its own copy as part of a Swing JTable component (GcodeTable.java), which stores even more metadata!!
4) (sigh) The GUI console may also keep a copy of this data, which would continue to grow as the operation is executed causing the application performance to degrade even further.

Basically a solution would need to address each of the above items to work in constant memory:

1) When commands are preprocessed they are built up into a collection of GcodeCommand objects. This could just as easily be piped into a new gcode file with the preprocessed data.

2) The visualizer is extremely naive in its implementation. I'm not sure how it could work in constant memory, but the memory usage could be optimized to do things like merge together tiny arc segments or line segments.

3) The data used by the JTable could be written to disk and looked up as needed. I'm sure there are a lot of options for this. One [possible resource](http://stackoverflow.com/questions/934927/efficient-way-to-implement-paged-sql-result-table-in-a-java-desktop-application), and [another](http://www.coderanch.com/t/345383/GUI/java/JTable-Paging).

4) The console output can be logged to a file, at which point there is probably a way to make the JTextArea discard old messages. A quick google turned up [this](http://compgroups.net/comp.lang.java.gui/buffered-jtextarea-or-equivalent/332460) which might be a solution.

winder:
Some issues this addresses #301, #244, #168

winder:
Possible solution for a disk backed JTable: [PersistentTableModel](https://github.com/Elblonko/kepler/blob/master/util/src/util/PersistentTableModel.java)

rugbymauri:
Is there really a need for the JTable? 
logging to the console, would be enough.

Or even the visualizer? I never need the visulizer.

winder:
The one nice thing you get from the jtable is that it matches up the
response with the command. But still I was thinking the same thing.

For the visualizer a lot of people use it as a preview tool. Maybe I'll
just add a warning when visualizing large files.

philreindl:
[...]

PR #326, opened 2016-01-09
Macros are stored in a collection

philreindl:
Move the macros into a collection.  The idea here is to set us up for having an arbitrary number of macros, rather than just 5.  The macro name is stored in the json, and can be used as the macro button label.  The description is stored, and is displayed as hover text.  Also, I moved the nightly build warning into the settings as well.

The changes in this pull request don't directly contribute much, but are setting the stage for better things to come.

### DIFF
--- a/src/com/willwinder/universalgcodesender/MainWindow.java
+++ b/src/com/willwinder/universalgcodesender/MainWindow.java
@@ -259,48 +259,6 @@ implements KeyListener, ControllerListener, ControlStateListener {
         });
     }
 
-
-
-    private void customGcodeText1ActionPerformed(java.awt.event.ActionEvent evt) {
-
-    }
-
-    private void customGcodeText2ActionPerformed(java.awt.event.ActionEvent evt) {
-
-    }
-
-    private void customGcodeText3ActionPerformed(java.awt.event.ActionEvent evt) {
-
-    }
-
-    private void customGcodeText4ActionPerformed(java.awt.event.ActionEvent evt) {
-
-    }
-
-    private void customGcodeText5ActionPerformed(java.awt.event.ActionEvent evt) {
-
-    }
-
-    private void customGcodeButton1ActionPerformed(ActionEvent evt) {
-
-    }
-
-    private void customGcodeButton2ActionPerformed(ActionEvent evt) {
-
-    }
-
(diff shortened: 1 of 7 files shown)


src/com/willwinder/universalgcodesender/MainWindow.java
```diff
@@ -2351,6 +2379,7 @@ private void setLocalLabels() {
         this.baudLabel.setText(Localization.getString("mainWindow.swing.baudLabel"));
         this.browseButton.setText(Localization.getString("mainWindow.swing.browseButton"));
         this.cancelButton.setText(Localization.getString("mainWindow.swing.cancelButton"));
```
winder: I spy some debug code!

winder:
This looks like a great step forward - I especially appreciate that you took the time to migrate legacy macros.

winder:
I'm in the final stages of finishing #302, so I'm going to hold off on merging these changes until I get that branch stabilized. But don't let that stop you with the next few steps, getting rid of all those duplicate functions will be nice.

philreindl:
Macro page is now rendered based on the macros that exist in the settings.  There is an extra, empty row to allow inserting new macros.  Since there's no add button, you need to close UGS and restart to get a new blank row.  Doesn't look like the panel is resizable, so (at least on my screen) you are currently limited to 8 macros.

It looks like the UI was originally built in NetBeans.  Is that the plan going forward?  Or is the ui code being manually maintained now?

winder:
@philreindl Awesome! Most of the UI code is generated with Netbeans, including some of the code you changed here (there are block comments pointing out the sections that are auto-generated). I'm not planning to change this.

The way I'm handling custom GUI code by putting it into a subclassed component then adding that component with the Netbeans GUI builder. I'm hoping to finish the constant-memory work this week (fixing up the last of the unit tests), I'm happy to pull the GUI changes into something the Netbeans builder can use afterwards.

philreindl:
Yeah I was afraid of that.  What had me questioning whether it was still in use is that the fonts and spacing don't match what I see in Netbeans.  Looks like there's some extra style being done outside the gui builder.  If you don't mind patching this up to be more netbeans friendly, that would be great.

philreindl:
@winder OK I took a crack at adding a new MacroPanel component, which is installed to a JScrollPane via netbeans GUI builder.  It all appears to be working.

PR #329, opened 2016-01-16
Resolved #275 - Restore parser modal state after jogging.

### LINKED ISSUE #275: Machine motion leave machine in incremental mode
When jogging the machine in the machine control panel, the GRBL status is left in incremental mode via the G91 command.  I suggest the machine be returned to the mode it was in prior to the jog command.  My machine is just about always in absolute mode.

Currently:
G91 X 1

Suggestion:
G91 X 1 G90

philreindl:
@winder Here's another one to review.  Issue a $G command on startup, tracking the units and movement mode.  After every jog, restore these values.  
edit:
I moved things around a bit.  Now, instead of sending a gcode string, the sending code first gets a GcodeCommand object.  The GcodeCommand object can track whether it represents a command that should overwrite the modal state or not.  Jogs set that value so that the state is restored, "normal" gcode commands do not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ tar.gz
 .idea
 *.iml
 dependency-reduced-pom.xml
+/UGSPlatform/dist/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: java
+
+jdk:
+  - oraclejdk8
+
+after_success:
+  - codecov
+  - mvn package assembly:assembly
+  - mkdir -p UGSPlatform/UGSLib/release/modules/ext/ugs/
+  - cp target/UniversalGcodeSender.jar "UGSPlatform/UGSLib/release/modules/ext/ugs/Universal G-Code Sender.jar"
+  - cd UGSPlatform
+  - ant build-zip
+
+before_script:
+  - pip install --user codecov
+
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Downloads
 
 <b>Stable builds</b>
 <br />
-[1.0.9](http://bit.ly/1M6z2ys) - The last stable build before 2.0 work?
+[1.0.9](http://bit.ly/1M6z2ys)
 <br />
 [1.0.8](http://bit.ly/1BSKon6)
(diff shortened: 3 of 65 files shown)


philreindl:
Working on getting the changes from master merged.  Hope to have it done tomorrow.

winder:
Sounds good, sorry about the code bomb.

philreindl:
No problem, I figured merge conflicts were likely.

philreindl:
@winder This is close, I have got a bit of work to do to figure out how the unit tests work.  The first issue was that my changes induce a $G on startup to get the current parser modal state.  The unit tests were not happy due to the serial connection having not been properly initialized.  In real life this would never be an issue of course.  I moved moved things around so that the AbstractController no longer directly tracks whether the serial port is open.  Instead, it delegates to the Connection.  Now, easy mock is mad because there's an unexpected method call being made in AbstractController where this delegation is happening.  

Before this is ready, I'll need to figure out how to tell easy mock what method calls to expect.  Not sure where that's happening yet.

Nice job getting automated tests running, btw!

philreindl:
OK, tests are passing, I think we're good to go.  There are a couple of tests I marked to ignore since I don't believe they are relevant any longer.  The tests were checking the isCommOpen flag, which got refactored to check the Communication class.  So the effect is we're now testing whatever we mock it out to be.  

The $G that is issued on startup also broke a bunch of tests.  In order to send the $G, the comm port must be open.  So tests that weren't opening the comm port or were expecting the comm port to be closed were broken.  I attempted to fix them all up.

winder:
@philreindl Thanks a lot for the work here - especially figuring out the tests. I think the next step is to rip out the units handling code that I have in the GUI backend and use this instead.

winder:
For example in GUIBackend.java the `commandComplete` function is setting the unitsCode variable twice. Also `GrblUtils.getUnitsFromStatusString` has got to go (GRBL sends out 3 fractional digits for MM and 4 for inches, using that felt like a hack).

I'll need to think about this a bit more, maybe the first step is to remove the units variable from GUIBackend and see where the cards fall.

PR #349, opened 2016-02-18
Branding

rugbymauri:
Added some Branding to the UGS Platform


winder:
Thank you this is great!

PR #352, opened 2016-02-21
Macros Window for UGS Platform

rugbymauri:
Hi Will i created a Macros Module for the UGS Platform. Macros can be edited and will be stored in a file. Each macro will add a Button with the Macro.


rugbymauri:
btw. i used http://glyphicons.com/ for the Icons

winder:
I haven't run the code yet, but I noticed you added a new property file for
the macros. You should be able to access the macros in the classic GUI via
the Settings object. You can get it from the CentralLookup.

I'll take a closer look at the PR later today, I've got a busy morning.

What was your impression after building a plugin? Did you find the tutorial
I wrote?

rugbymauri:
Yes i used the turorial. i just strait forward. Great

i looked the the Settingd in the Central Lookup. But for my macros (it more an "CustomButtons") i need two fields: a title and the Code. in the Settings i only can store the Gcode.

winder:
@rugbymauri There is actually a name/description attached to each Macros stored in the settings now. Take a look at [Macro.java](https://github.com/philreindl/Universal-G-Code-Sender/blob/fd9241eb428418e2215d4be420f4034ea3f635b4/src/com/willwinder/universalgcodesender/types/Macro.java) and [Settings.java:updateMacro/getMacro](https://github.com/philreindl/Universal-G-Code-Sender/blob/fd9241eb428418e2215d4be420f4034ea3f635b4/src/com/willwinder/universalgcodesender/utils/Settings.java#L332).

@philreindl cleaned up all that code in #326

PR #366, opened 2016-03-19
Move file panel to left under machine status

philreindl:
Moved the file panel to the left column under the machine status.  The console is made narrower, but it seems to be mostly wasted space anyhow since the gcode commands are limited in length.

image

I'd like to reduce the file control buttons.  Send and Pause can share (send changes to pause while sending.)  Browse and Cancel can share.  (Cancel while sending, browse while stopped.)  Visualize can move to machine control.  I'm not sure what the save button is for.  It doesn't appear to create a valid gcode file.

The next step would be to pull the jog panel out of Machine Control.  Machine control would then get a big grid area for the custom macros.


winder:
In general I think this layout looks a lot better, just a couple comments:

- It looks like the row numbers are offset from their labels in the new box.
- Do you think having the browse button so far away from the file box could be confusing?
- In the platform build I've done away with the file browse button all together in favor for an "Open Gcode File..." menu option, that might be an option here too.

philreindl:
I'll get the text alignment fixed.  When the application first starts up, all the file buttons except "browse" are grayed out.  I'd eventually like to get tool tips on everything which should help.  But your options are pretty limited, so IMO the current design (even without these changes) is usable enough.  

My preference would be to avoid menu options for normal workflow commands.  My tendency would be to reserve the menu controls for configuration settings.

PR #368, opened 2016-03-20
Move jog panel to left, making room for macro buttons

philreindl:
image


faultylee:
Just wondering if the panel will be scroll-able? I'm running this on an EEE PC@1024x600. I've not tested the latest build yet. Might be helpful also if the panel is collapsible.

winder:
@faultylee thanks for the feedback

philreindl:
There's a conversation about this going on over at #374.

PR #373, opened 2016-03-24
Add split panes everywhere.  Resolves #317

### LINKED ISSUE #317: Why UGS of machine status in windows 7 & 8.1 ,Z position is covered by half?
machinstate
Z position is covered by half

philreindl:
image

### DIFF
--- a/src/com/willwinder/universalgcodesender/MainWindow.form
+++ b/src/com/willwinder/universalgcodesender/MainWindow.form
@@ -567,230 +567,21 @@
 
               <Layout class="org.netbeans.modules.form.compat2.layouts.support.JSplitPaneSupportLayout"/>
               <SubComponents>
-                <Container class="javax.swing.JPanel" name="keyboardMovementPanel">
-                  <Properties>
-                    <Property name="minimumSize" type="java.awt.Dimension" editor="org.netbeans.beaninfo.editors.DimensionEditor">
-                      <Dimension value="[10, 10]"/>
-                    </Property>
-                    <Property name="preferredSize" type="java.awt.Dimension" editor="org.netbeans.beaninfo.editors.DimensionEditor">
-                      <Dimension value="[247, 220]"/>
-                    </Property>
-                  </Properties>
-                  <Constraints>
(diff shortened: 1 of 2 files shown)


winder:
All the frames on the left are a fixed size, what is this for?

philreindl:
No, they can be resized, but it's not quite as nice as I'd like.  You have to do the middle first to make room for the top.  Let me mess with it some more.  Had a hard time getting the default layout to look right.

philreindl:
Give this one a go.  I had been using minimum size where I should have been using preferred size.  Minimums are much smaller now, preferred size actual sets the default sizing for everything.

philreindl:
Ok for real this time.  Just realized the file and jog panels got switched.

winder:
The merge conflict is changing `MacroPanel(settings, this)` to `MacroPanel(settings, backend)`. All that auto generated code got messed up, just swap those two constructor calls in MainWindow.java/MainWindow.form.

I used your new macro panel in the platform config menu, and am planning to do the new MacroActionPanel next.

winder:
So I tried this out, and see the split panes everywhere. The one between the console / tabs is pretty cool, especially now that those macro widgets can grow.

The ones on the left (and between the right/left) I'm not so sure about though. If the only reason for them is to fix #317 I don't really like it since the content in these panes is almost completely fixed size - if I were to ever drag them it would be a mistake.

winder:
I also modified `updateControls` to remove the `this.macroPanel.updateCustomGcodeControls(*);` and `this.macroActionPanel.updateCustomGcodeControls(*);` calls. Those panels keep themselves updated now.

philreindl:
I'll pull changes and get that fixed up.  Busy weekend, so it might be a few days.  (On the other hand, might be a good excuse to hide for an hour.)  As to the sliders on the left, I agree they probably shouldn't be necessary.  But given the variety of environments this is running in, it seems like a relatively easy fix for issues like #317 which to my knowledge hasn't been reproduced by anybody.

philreindl:
Should be good to go now.

winder:
I still don't like the split panes on the left side. But moving them around a bit does show just how messed up the GUI builder frames are though. Whatever you did with the new file box seems to prevent the height from being too small.

PR #377, opened 2016-03-30
Edge

philreindl:
Bring edge up to date with master.


winder:
Unfortunately, I think this is going to trigger a build, and update the nightly build file. Lets see.

philreindl:
Dang, I was hoping that the edge branch would get a build based on the edge branch, which wouldn't impact the master branch's build.  Looks like the nightly build artifacts on both are the same.

winder:
I was thinking we could add a menu called "experimental" with an option to launch the new UI, that way your work would stay more visible and the current UI would still work.

philreindl:
I'd rather not go down that route.  The advantage of it will work without messing with the build system is offset by the added complexity of setting it all up in the program.  IMO, the ideal way to handle it would be to continue working on MainWindow, but use a branch (that's what they're for.)  Considerably clunkier would be to copy MainWindow to ExperimentalMain and go to town.  Now refactoring is more dangerous since MainWindow and ExperimentalMain will be sharing a lot of code.  Both approaches suffer from not being able to generate builds without changes to the build system.

Another option would be to keep going the way we have been.  Realizing now that there's more traffic than I thought on the nightlies, I'll be more careful to not break anything.  :)  This is probably the best/most efficient approach.  We'll get feedback if things break.  I would think that people understand that's part of the deal with nightly builds.  Users can always go back to 1.0.9 if there's trouble.  Another factor in favor of this option is that I've got a significant work project looming which threatens my free time, so I may not be able to contribute as much for a while.

winder:
@philreindl I agree for the most part. The release process for UGS is a little unorthodox simply so I don't have to do releases as often (I probably should have done one a while ago, since the "constant memory" feature is pretty important).

Most of the stuff left in MainWindow.java is still there because it's the only thing left using it, so any refactoring you do should be pretty isolated. Let me setup a demo of what I have in mind for my "considerably clunkier" approach - I'll use the edge branch. If you still aren't convinced, using a long running branch, or just being a little more cautious in master is fine. Mainly I'd just like to make sure there's a spot for you to go to town on the UI and get feedback from the community.

Also, could you drop me an email? I had a couple questions. My email is in my github profile.

PR #381, opened 2016-04-03
Updates to experimental UI

philreindl:
Connection Panel and Machine Status panel are both MigLayout components.
Use the File Status Panel.


winder:
One of the maven repositories is down which is why the build failed, don't worry about the build failure.

I did build script refactoring in the edge branch and forgot to remove the ExperimentalWindow. I can hook the experimental window back up to the legacy UI if this works for you, or you can keep working on MainWindow.

philreindl:
Yeah, the build failure looked like something that wasn't related to what I had changed.  So I figured there was something else going on.  

I have been just working on the ExperimentalWindow file, which is sitting next to MainWindow.  I'm fine keeping it as is for now.  We could add it as another build artifact or hook it up to MainWindow.  But I think your point of keeping MainWindow fairly stable is reasonable.  This PR was mostly just to keep your repo current with my changes so that merge conflicts are less likely as we go.  I'll reopen this PR when I've got connection panel and machine status merged together.

PR #389, opened 2016-04-11
Console Panel

philreindl:
Console panel is a UI control.  Removed the command table and the bottom tab panel from the experimental UI.


winder:
I like where you're heading with all this, but the spacing of things is looking really wonky. Does it do this on your system?

screen shot 2016-04-11 at 8 37 12 am

philreindl:
My first order of business is to get all the UI components to be handled with miglayout.  From there, I will dial in the layout.  The blend of netbeans managed layout and miglout layout isn't the best.  I think that the trouble with the components on the left is that there's a mix of hard-coded (non-miglayout managed) and miglout managed sizes.  The macro action panel needs some love.  I don't like how it vertically spaces everything.  That said, it looks better on my machine than yours.  ;)

PR #393, opened 2016-04-17
Conversion to mig layout continues

philreindl:
Most of the UI is managed in MigLayout.  All the panels on the left got merged, which should help considerably with the layout.  Visualizer is temporarily unavailable, will add that in as a tab next.


philreindl:
Visualizer has been moved to a tab

PR #394, opened 2016-04-24
Experimental UI changes

philreindl:
Add z step spinner.  Better increment/decrement logic.  Pendant menu is a control.


philreindl:
@rugbymauri Question for you.. Do you primarily use inches or mm for the jog?  The reason I'm asking is that the 1 unit increment makes much more sense in inches than mm.  If I wanted to go up to a 50mm jog, then that's a lot of clicking if you have to go by one.  That was the motivation behind changing your algorithm.  I'm realizing that it might not be a nice change for those who run in inches.

jahnj0584:
can we program it like the oven timer, where after you get to 10 from 0, it starts jumping by 10 instead of 1's? 
So 1 2 3 4 ... 9 10 20 30 40 etc

philreindl:
Yeah, that's the idea I was going with..

.001, .002... .009, .01, .02,...09, .1, .2, .. .9, 1, 2, .. 9, 10, 20, .. 90, 100, 200, .. 900, 1000

The 100 steps might be a bit much, I believe that  @rugbymauri is suggesting that the 10 steps might be too much as well.

jahnj0584:
if youre programming it in, surely it cant be too hard to add it as a menu option to choose our scales?

philreindl:
Too hard?  No, not really.  And if somebody wants to add in customized step limits, I certainly wouldn't object.  In my view, the increment/decrement/multiply/divide buttons are all falling short of the real nice to have feature.  Ideally, we'd have the ability to define buttons that set the jog to something in particular.  I generally find that I want to either jog big (50-100mm), medium (~10mm), or small (<=1mm).  For that to be really slick, you'd need to be able to define hotkeys for all the custom buttons.  Custom hotkeys is the next major effort I intend to tackle after the primary UI work is done.

rugbymauri:
Hi

i'm using "mm". And i have a separate keypad for jogging the machine, i not often use the mouse.
my default setting is 1mm for the x/y axis, because 1mm it's "realtime" on my machine (if keep the key pressed it moves until i release the key, without buffering to much moves)
i think also about this big/medium/small steps, but the problem is the GUI and the leak of key on the keypad. A solution could be to configure a custom steps value-list and switch between them by +/-, perhaps one for inches and one for millimeters.
One of my next project is to build a custom Jog-Keypad with more key than a regular Keypad (using a tensy-controller), so custom Hot-Keys would be a great improvment!!!

winder:
@philreindl we're going to need to figure out how hotkeys in the swing GUI coexist with with hotkeys in the platform GUI, some of these features are already implemented there.

philreindl:
> A solution could be to configure a custom steps value-list and switch between them by +/-

I was thinking about it last night, and taking a step back, I think that's the same solution I came up with.  The more I think about it, the better I like it.  

As to the jogs "leaking", you are referring to the fact that jog command (like all gcode commands) are buffered, right?  We could probably make some changes in the backed to not accept new jogs until the buffer is cleared.  Not sure how much work that would be or whether it would be desirable in the end anyway.  Probably worth investigation though.

winder:
There is a "single step mode" which can be enabled which prevents multiple commands from being sent at a time, but GRBL has optimizations to prevent decelerating between each jog command so this probably wouldn't work well.

Another option is to send a single large movement while the arrow key is held down followed by a pause/soft-reset when the arrow is released. This wouldn't work with all controllers/versions of grbl though so it would need special checks before enabling the logic.

faultylee:
Slightly off topic, with regards to the special checks, I saw an implementation from FinePrint, which has a wizard that prints a few pages, and based upon the print out, decide how to handle duplex printing. 

Maybe something along that line for UGS, where a few test is carried out and a recommended settings is presented to the user. That should help cater to the different controllers/versions and maybe even configuration of grbl

PR #396, opened 2016-04-26
Fixes to the jog spinners

philreindl:
Experimental UI jog spinners now increment/decrement/muliply/divide properly


winder:
@philreindl the `commitEdit()` call is or the case where someone is typing in a step size and then clicks a jog button. In that situation the old value is used rather than the newly edited value, so I think that code is still needed.

philreindl:
@winder Changes are in

PR #514, opened 2016-10-22
GRBL v1.1 Jog Mode + UI Changes

winder:
Lots of changes to consolidate duplicate jogging code in the core and platform projects. They are now both using the same UI element as well.

Once the code was cleaned up, support for the new `$J=` mode was added. It is only enabled with supported versions of GRBL ( >= v1.1).

The UI now lets you use separate step sizes for XY and Z, and lets you configure the feed rate.

jog_panel

jog_mode


winder:
I'm trying out making a pull request for larger changes like this.

jahnj0584:
Can you remove the box with the " in it and instead use it for" common jog increments" like 0.1/1/10/100?

winder:
@jahnj0584 I was thinking about doing something like that, but it might make the jog widget a lot larger since it would potentially need 3 sets of the common increment selectors. Maybe I can make a separate widget to try this out. Here are a couple other UIs that I'm looking at for inspiration. With the exception of pronterface (which I don't like) all of these look larger than the jog panel in UGS:

grbl-panel
pronterface
repetier-host-osx
repetier-host-windows

jahnj0584:
I wouldn't use feed rate there, and then put the selection choices under
the Z buttons so it's the same height as the Y-/+ buttonz

Encarnacao:
whats the mm and the "" new square, i notice now when first connect machine im not able to use keyboard for jog

kardinal7:
is it possible to bind some keys to use for jogging. like in the classic UI up/down/left/right/PageDown/PageUp. 
I think it's only missing the homing and the XY step size and the Z step size.
I was checking and the possibility to jog the machine with a game controller is great.

winder:
@kardinal7 yes it is! In the preferences under keymap:
 
screen shot 2017-02-10 at 6 41 25 pm

kardinal7:
That's awesome. Thanks, I didn't check the options.... in windows it appears the standard shortcuts with meta+up. What is the META key?

winder:
@kardinal7 It might be the windows button, I'm not sure though. Looking at the netbeans documentation it mentions `Do not use the Meta key (the Command key on the Macintosh platform) for a keyboard shortcut, except as an alternate for Control. It is not available on some target platforms`

So maybe I should change the default to ctrl. You should be able to set it to whatever you like, good luck with the joystick.

jahnj0584:
All you need is xpadder

https://youtu.be/Y-erKT4zRlY

PR #611, opened 2017-03-30
Update MacroService.java

linno60:
copywrite -> copyright


winder:
Wow, every file is like that since the beginning of this project.

Issue #619, opened 2017-04-13
Customizing the action toolbar throws errors if switching to other locales


joacimb:
If I switch to another language the toolbar customization tool throws an error. The only way to restore UGS is to manually remove the action instance files and restart. 

To reproduce:
1. Open UGS and change language to German then restart UGS
1. Choose to customize the toolbar ![screen shot 2017-04-13 at 06 49 31](https://cloud.githubusercontent.com/assets/8962024/24990617/9f6a8610-2015-11e7-9ec9-cbd717e3a408.png)
1. An error is displayed that the action isn't translated: ![screen shot 2017-04-13 at 06 52 08](https://cloud.githubusercontent.com/assets/8962024/24990665/dd282912-2015-11e7-86ee-cd4dea3ae128.png)
1. The customization window now shows all actions from UGS, but they can not be used: ![screen shot 2017-04-13 at 06 54 59](https://cloud.githubusercontent.com/assets/8962024/24990756/782c767a-2016-11e7-9b6f-7edd96ceb3ab.png)
1. Files for action instances are created for both languages:
![screen shot 2017-04-13 at 06 58 54](https://cloud.githubusercontent.com/assets/8962024/24990807/c72219ba-2016-11e7-94b0-38d7cf053ec0.png)

The customization tool can not be used right now and requires more tending. But fixing this bug could be a first step and should prevent unrecoverable errors for the users.

joacimb:
I can't recreate the issue after the fix. I've tried switching to several languages on different platforms. I think we can close the issue.

PR #620, opened 2017-04-13
Added unique identifiers for all actions, issue #619

joacimb:
Fixes the error message when opening the toolbar customization tool after the locale has been switched. Issue #619

The identifiers unfortunately needs the addition of a magic string. Decide if you think this is the right solution. :)

This could maybe be solved by using the action class name. But this requires that the action only has one usage, as opposed to OverrideAction for example which has many actions built into it.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/i18n/Localization.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/i18n/Localization.java
@@ -41,7 +41,6 @@ public class Localization {
     private static ResourceBundle english = null;
 
     private static int englishKeyCount = 0;
-    private static Locale locale = null;
     private static String region = null;
 
     /**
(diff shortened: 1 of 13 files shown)


winder:
@joacimb wow, thanks for digging into this. The programmatic way I wanted to handle localization is non-standard for Netbeans Platform applications... which led to the piece of code you found here.

I'll need to do a little bit of testing, but this looks like a good solution.

ugs-core/src/com/willwinder/universalgcodesender/i18n/Localization.java
```diff
@@ -56,18 +55,20 @@ synchronized public static boolean initialize(String language) {
...
-        locale = new Locale(language, region);
+        Locale locale = new Locale(language, region);
+        Locale.setDefault(locale);
         bundle = ResourceBundle.getBundle("resources.MessagesBundle", locale);
         return getKeyCount(bundle) >= getEnglishKeyCount();
     }
 
     public static String loadedLocale() {
-        return locale + "";
+        return Locale.getDefault() + "";
```
winder: This isn't going to have the expected behavior, the `getDefault` call needs to go to the location where `initialize` is being called. I load some GRBL code number translation files with this (even though GRBL doesn't have localizations yet).

joacimb: Hmm, I could be missing something, but I thought that by storing the selected locale using "Locale.setDefault" in the initialize-method would be the same as storing it as a local member?

winder: Thanks, I understand now, I thought this was changing the UGS localization based on the system locale. I didn't realize the `Locale` defaults worked this way.

winder:
I tried out the Locale changes but it didn't seem to localize the NetBeans menus, are those bundles separate downloads?

joacimb:
I have not found any translation bundles for other languages for the standard platform. I tried overriding translations for existing menu items using this in `layers.xml`. Some menu items can be translated while others (eg. Help) won't for some reason...
```xml
<folder name="File">
    <attr name="SystemFileSystem.localizingBundle" stringvalue="resources.MessagesBundle"/>
    <attr name="displayName" bundlevalue="#file"/>
</folder>
```

Issue #621, opened 2017-04-15
Add possibility to customizing the toolbar


joacimb:
I've been poking around in the code trying to get the toolbar customization working to make UGS more suitable to run on a  computer with a low resolution touch screen. I believe that by making the platform more customizable, UGS can support more platforms and might be easier to extend. 

Take the toolbar panel for handling the GRBL connection for instance. If we break it apart we could hide the settings for firmware or baud-rate as it's something you seldom change. But for some platforms this might be a requirement and would still be supported.

Also, if we try to utilize more functionality from the NetBeans platform we would, as an example, be less dependent on our own action registration procedure and could also get localization and icon-handling by the platform. 

In the picture below i've experimented with the connection and run actions. I've removed the StartStopPauseToolbar and removed the connect-action from the ConnectionToolbar:

![screen shot 2017-04-15 at 12 42 12](https://cloud.githubusercontent.com/assets/8962024/25063834/6d02caa4-21ee-11e7-8d78-7cc11d632adf.png)

So I would like to try to:

1. Move some of the common actions from ugs-core to ugs-platform/ugs-core and annotate them with NetBeans action annotations. This also requires the constructor that takes the backend parameter to be removed and let them use the CentralLookup instead. The actions also needs to be removed from the RunActionService so that they are not duplicated.
2. Remove ConnectionToolbar and replace it with smaller actions
3. Remove StartStopPauseToolbar and replace it with smaller actions

What do you think?

winder:
One of the main reasons so much of the GUI code is in ugs-core is to allow me to support the classic GUI as well as the platform. I would like to use more of the NetBeans platform, and this looks like a big improvement to what I'm currently using, but it would need to be done in a backward compatible way.

About localization, I really didn't like the way it was done with NetBeans - every package needed a separate localization file and couldn't figure out a way to reuse the existing localizations with the platform interface. I'm open to suggestions, as the current approach isn't great (The keyboard shortcut menu cannot be localized). I didn't know about the toolbar customization feature, who knows what other features are broken by the approach I'm using.

Specific to your suggestions:
1. I really like the idea of leveraging more NetBeans platform features, but since I'm still supporting the two interfaces I've been using facades to register standard swing code into the platform rather than migrating it completely. You can see most of the `*TopComponent.java` classes, like `JogControlTopComponent` for how I've been using the NetBeans annotations to register standard swing components.  Doing this with a `StopAction` which wraps `Stop` for instance seems like a great idea.

[...]

PR #624, opened 2017-04-17
Toolbar customization. Issue #621

joacimb:
The toolbar is now splitted into several panels and are now customizable.


ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/toolbars/Bundle.properties
```diff
@@ -0,0 +1,4 @@
+mainWindow.swing.firmware.toolbarTitle = Firmware selection toolbar
```
winder: These can be put in `LocalizingService.java` to statically load the translations from the properties files instead of having them localized in two places.

winder:
I took a quick pass and it looks good! Using the annotations to configure these should really help with integration. I'll need to spend some more time on a full review later.

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/toolbars/Bundle.properties
joacimb: Unfortunatly I cant use the localizing service for the action display name (the annotation requires a static final variable). I was hoping that it could be assigned dynamically in the constructor and it will work everywhere, except for in the toolbar customization... :(
```java
@ActionRegistration(
        iconBase = PauseAction.ICON_BASE,
        displayName = "<Not localized> PauseAction",
        lazy = false)

    ...

    public PauseAction() {
        ...
        putValue(NAME, LocalizingService.PauseTitle);
        putValue("displayName", LocalizingService.PauseTitle);
    }
```
screen shot 2017-04-17 at 20 01 58

By giving a resource bundle it will work if the bundle exist in the same module with the key. But this leads us to the same problem with multiple bundle files. 
```java
@ActionRegistration(
        iconBase = PauseAction.ICON_BASE,
        displayName = "resources.MessagesBundle#mainWindow.swing.pauseButton",
        lazy = false)
```

I'll move the definitions to the ´LocalizingService.java´. If we want to use the correct display name for the actions I think we need to consider having local bundles in the package with a default translation. All overrides can still be done in the central MessagesBundle. But this is for you to decide. :)

winder: @joacimb I also tried to set everything in the constructor and started several discussions on the NetBeans mailing list about different localization subjects like this. I think the annotations are package scanned by a separate service to register everything, so it really doesn't work this way. So everything needs to be initialized in a static final variable so that it can be annotated outside the class, which is what I use `LocalizingService.java`. It declares a bunch of static final strings.  In `JogControlTopComponent` for instance is this:
```
@ActionID(category = LocalizingService.JogControlCategory, id = LocalizingService.JogControlActionId)
@ActionReference(path = LocalizingService.JogControlWindowPath)
```

Which properly registers the different modules into the menu. It would be nice if annotations could be set dynamically, but I wasn't able to find a way to do it.

Looking at it a little more, the toolbar instances might need to be updated in the `LocalizingService` constructor the same way menu items are updated.

joacimb: What I meant was that I can not use LocalizingService to translate a title like this:
```java 
public final static String PauseTitle = Localization.getString("mainWindow.swing.pauseButton", lang);
```

And use it in the registration like this because it isn't a constant:
```java
@ActionRegistration(
        iconBase = PauseAction.ICON_BASE,
        displayName = LocalizingService.PauseTitle,
        lazy = false)
```

However, I've moved the key constant to the localizing service and reference that as bundle key. But this requires that there is a bundle file in the same directory.

joacimb: I've been experimenting more with this and in the way the LocalizingService is implemented it can't be used together with the ActionRegistration-annotation unless we make some compromises.

As I see it there are three options:
1. Let the actions be registered using the netbeans api which will require an extra localization bundle with default texts (which can be overrided with the MessageBundles) in the same way I've experimented with. The greatest benefit that I see it is that we will get autodiscovery of actions and their configuration (no need to manually register them like in MacroService, JogActionService, RunActionService). We will also get a lot less code to maintain and get more features like scaling icons and the translation of action names etc. The disadvantages is that the configuration is more spread out and if we wish to disable an action we need to override the configuration in layers.xml. Also the documentation for Netbeans platform is not the best (I needed a couple of days to figure out how all this worked...).
2. Continue to use ActionRegistrationService and let all translations be handled by LocalizationService. The greatest benefit is that all actions are registered in one place, one configuration to rule them all and we are in the control of the process. The greatest disadvantage is that we need to register our actions through services that we need to create. We would also need (fairly) complex logic to determine that an actions no longer exist and remove its instance file from the file system.
3. We mix them. Scratch the changes that I've experimented with and let the core functionality use the action registration and localization as it's implemented today. Plugin developers can use netbeans platform api and use it's standard autodiscovery functionality, etc. The disadvantages with this I think is that the way we implement things would be inconsistent, there would be no good recipe in how to create a plugin. Also, the only way to disable core functionality would be to make code changes.

winder:
Thanks for the feedback, I've been doing this implementation alone so a
second opinion is incredibly useful. You're probably right about using more
of the nerve and API - the stale instance objects specifically are really
annoying.

For this PR I think your changes are probably fine to merge as-is, but I
still haven't had a chance to pull them down and try it out myself.  Your
contribution to the platform code is really exciting for me, so sorry if
I'm giving you a bad impression by letting the PR sit for so long!

joacimb:
No worries, take your time! It's important it gets right and that it's the best solution for the project which I'm sure you are the best person to decide. I'm just trying to make sure we have all the facts to easier make a decision.

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/StopAction.java
```diff
@@ -0,0 +1,88 @@
...
+
+    // Icons: http://www.customicondesign.com/free-icons/flatastic-icon-set/
+    public static final String ICON_BASE = "resources/icons/stop.png";
+
+    private BackendAPI backend;
+
+    public StopAction() {
+        this.backend = CentralLookup.getDefault().lookup(BackendAPI.class);
+        this.backend.addUGSEventListener(this);
+
```
winder: @joacimb It looks like these values cover everything needed for overriding the localization. I don't see any negative effect to removing the bundle reference in the ActionRegistration annotation:
```
+@ActionRegistration(
+        displayName = "StopAction", // not used - required value
+        lazy = false)
```

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/toolbars/Bundle.properties
winder: I agree that the NetBeans documentation isn't great once you try to extend the default behavior.  There is a lot of great example-driven documentation though, which is nice (when there is an example close to what you need). The developers are pretty active and I've gotten decent support on the mailing list. [Here is a good example :)](http://wiki.netbeans.org/DevFaqActionsAddAtRuntime) 

I finally see what you were originally saying about the need for `displayName` to be a constant. I thought that the TopComponents were using derived strings, but actually, they call a `setName` method to do the same thing as the `putValue(NAME, ...)` method you found.

There are two main reasons I wanted the `ActionRegistrationService`:
1. Macros. This is the original reason. These are completely dynamic, so there is no way to register them with individual classes.
2. Repetitive actions. The `JogActionService` has almost 20 actions that sometimes differ by just a few characters. Should these be separate classes? I already had the service available for macros, so I extended it to add the default keyboard shortcuts.

I think that the stale instance objects are the most compelling reason for your first option, that alone is enough of a reason that I really like the approach you have here. The extra bundle files it seems is a non-issue (they aren't even really needed for the code in this PR). Maybe having 20 small classes with descriptive annotations is better than 20 calls to the `ActionRegistrationService` with a bunch of cryptic arguments.

It still needs to be a hybrid approach though. I don't see an alternative for Macros, and `ActionRegistrationService.createAndLocalizeFullMenu` needs to be called somewhere to localize the menu hierarchy.

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/StopAction.java
joacimb: It's used by NetBeans toolbar customization tool:
screen shot 2017-04-23 at 08 05 52

It doesn't seem to use the object instance parameter, which seems logical because it would otherwise need to instantiate all actions available to the platform to get the name.

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/toolbars/Bundle.properties
joacimb: An alternative for the MacrosService could be to create an action that presents a dynamic menu. Something like this:

```java
package com.willwinder.ugs.nbp.core.actions;

import com.willwinder.ugs.nbp.lib.lookup.CentralLookup;
import com.willwinder.universalgcodesender.MacroHelper;
import com.willwinder.universalgcodesender.model.BackendAPI;
import com.willwinder.universalgcodesender.types.Macro;
import org.openide.awt.*;
import org.openide.util.Exceptions;
import org.openide.util.actions.Presenter;

import javax.swing.*;
import java.awt.event.ActionEvent;
import java.util.Collection;
import java.util.List;
import java.util.stream.Collectors;

@ActionID(
        id = "com.willwinder.ugs.nbp.core.actions.MacrosListAction",
        category = "Machine")
@ActionRegistration(
        lazy = false,
        displayName = "MacrosListAction")
@ActionReferences({
        @ActionReference(
                path = "Menu/Machine/Macros",
                position = 1000)
})
public class MacrosListAction extends AbstractAction implements DynamicMenuContent, Presenter.Popup {

    private final BackendAPI backend;

    public MacrosListAction() {
        this.backend = CentralLookup.getDefault().lookup(BackendAPI.class);
    }

    @Override
    public void actionPerformed(ActionEvent e) {
        // does nothing, this is a popup menu
    }

    @Override
    public JComponent[] getMenuPresenters() {
        return createMenu();
    }

    @Override
    public JComponent[] synchMenuPresenters(JComponent[] items) {
        return createMenu();
    }

    @Override
    public JMenuItem getPopupPresenter() {
        JMenu menu = new JMenu(this);
        JComponent[] menuItems = createMenu();
        for (JComponent item : menuItems) {
            menu.add(item);
        }
        return menu;
    }

    @Override
    public boolean isEnabled() {
        return backend.isConnected() && backend.isIdle();
    }

    private JComponent[] createMenu() {
        Collection<Macro> macros = backend.getSettings().getMacros();
        List<JComponent> actions = macros.stream()
                .map(this::makeMenuItem)
                .collect(Collectors.toList());

        return actions.toArray(new JComponent[actions.size()]);
    }

    private JComponent makeMenuItem(Macro macro) {
        JMenuItem item = new JMenuItem();
        Actions.connect(item, new MacroAction(backend, macro), false);
        return item;
    }

    private class MacroAction extends AbstractAction {
        private BackendAPI backend;
        private Macro macro;

        public MacroAction(BackendAPI backend, Macro macro) {
            this.backend = backend;
            this.macro = macro;
            putValue(NAME, macro.getName());
            putValue("menuText", macro.getName());
            putValue("displayName", macro.getName());
        }

        @Override
        public void actionPerformed(ActionEvent e) {
            try {
                if (macro != null && macro.getGcode() != null) {
                    MacroHelper.executeCustomGcode(macro.getGcode(), backend);
                }
            } catch (Exception ex) {
                Exceptions.printStackTrace(ex);
            }
        }

        @Override
        public boolean isEnabled() {
            return backend.isConnected() && backend.isIdle();
        }
    }
}
```

This should also utilize the menu translations if it's defined in the layers.xml.

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/StopAction.java
winder: Gotcha. I noticed in the manifest the localization bundle is specified like `OpenIDE-Module-Localizing-Bundle: com/willwinder/ugs/nbp/core/Bundle.properties`, I wonder if there's a way to modify it to use the existing `MessagesBundle.properties` files.

I did some experiments and wasn't able to so I'm asking on the NB forum: https://forums.netbeans.org/viewtopic.php?p=174898#174898

joacimb: Yep that should work, but there needs to be a default MessageBundle.properties in the same module, this will be overridden by the master MessagesBundles. I'll give it a go!

joacimb: It works but it seems that if I specify another bundle using `resources.MessagesBundle#mainWindow.swing.pauseButton` NetBeans will try to read the file using ISO-8859-1 instead of UTF-8...

winder: @joacimb that sounds really promising! I don't see any reason not to switch from UTF-8 to ISO-8859-1 if it makes this integration easier.

So you didn't need to change `OpenIDE-Module-Localizing-Bundle` in the manifest, just add `resources.MessagesBundle` before the # reference?

joacimb:
There is no need to modify the manifest, just to add `resources.MessagesBundle` before the # reference.

I tried converting the messages bundles to ISO-8859-1 then it works in the customization tool but not anywhere else in the GUI. =P

Soo... I've moved the small bundle files to a default messages bundle. This seems to work everywhere except in netbeans customization tool and should be good enough?

winder:
I want to keep experimenting, but like where this is going.

winder:
@joacimb I noticed this today: http://openjdk.java.net/jeps/226

I'm also wondering if I pulled in some of your code that I shouldn't have. I see the `ISO-8859-1` conversion stuff in master:

```
ugs-core/src/com/willwinder/universalgcodesender/i18n/Localization.java
91:            result = new String(val.getBytes("ISO-8859-1"), "UTF-8");
102:                result = new String(val.getBytes("ISO-8859-1"), "UTF-8");
```

Since POEditor writes these files, I'm not actually sure what the real encoding is.

joacimb:
Sorry for the late reply. 

Great that they have been working on this. I don't remember the different experiments I tried out to get the UTF-8 bundle file to be loaded, but the problem was that it wasn't working everywhere. I don't remember exactly where I saw the code, but I think that NetBeans loads the properties files specifically with ISO-8859-1.

The converting code was there from the beginning. But I've noticed several reports of translations missing giving blank texts in the UI. This is actually a bug in the different localization bundles where the key is defined with an empty text. I added a check if the string in the bundle was empty it will try to fetch the english translation.

PR #630, opened 2017-04-20
Fixed bug with the jog status line that didn't get updated

joacimb:
Fixed bug where the jog status line unit wasn't updated and that event for the changed setting was fired before the setting was stored.

screen shot 2017-04-20 at 22 07 18


winder:
Nice catch

vMeph:
nice, by the way couldnt the inch insted of showing "" just say inch?
sounds less confusing no?

Issue #625, opened 2017-04-19
Inconsistent "Return to Zero" results


cjbaar:
I am having difficulty getting useful and consistent results from "Return to Zero" function. Maybe I am not using it correctly, but I am not sure how else to accurately reposition the spindle after changing bits.

Sometimes the command just sends "G90 X0 Y0 Z0". When it does this, my arduino-based board processes the Z-axis first, so the bit ends up dragging across the work piece. Other times, it tries to raise the spindle first, but comes up with completely random and large values, such as "G90 Z12.7." My Z-axis only has about 3" of play, so it immediately shoots to the top and the motor starts grinding before I can react and hit the kill switch. I have even seen Z values sent over 4,800 inches.

I can write my own in the macros, I suppose... but is there a way to control what the "Return to Zero" function sends?

jahnj0584:
I sent this issue before. Are you on the latest version? A work around is
jogging the Z to its machine 0 first then clicking it

cjbaar:
I am using version 1.0.9, which appears to be the latest in the stable branch. I have tried manually moving the spindle close to home before clicking it, in order to prevent the bit from dragging too far. However, this doesn't help anything when UGS decides to send a "Z4896.3" command and raise the spindle 400 feet into the air.

jahnj0584:
Try with the platform version

winder:
@cjbaar those issues should be fixed in the nightly build version

cjbaar:
So far, much better with the platform version. Guess I'll stick with that from now on. Thanks for the input!

winder:
@cjbaar Thanks for the update!

cjbaar:
Actually, I was to quick to respond. Doing some more work this evening (using platform edition), and the return button issued:
>>> G90 G0 Z0
>>> G90 G0 X0 Y0
>>> G90 G0 Z0
Again, since the Z0 was issued first, I got a nice scratch across the top of the piece. I would just like to be able to edit what commands are sent so I know that they are consistent.
Thanks again for all your work!

joacimb:
I've experienced scratched surfaces by homing myself, mostly because my crappy home built CNC machine isn't absolutly leveled... =) 

I saw that the max position has previously been used but removed because of people experiencing problems with this feature. I'm thinking that we could instead use an optional setting for Z clearance distance to be used?

jahnj0584:
If you're scratching your Z is too low anyways.

winder:
[...]

PR #631, opened 2017-04-21
#625 I wrote two tests that confirms that the go-to-home button goes to Z0 first

joacimb:
I wrote two tests that confirms that the go-to-home button goes to Z0 first if the work Z-position is negative. This could help understanding the issue #625?


winder:
I think the problem was somehow related to homing. I was trying to record the maximum Z-height that had been seen instead of requiring an extra clearance configuration. The bug people experienced was that the number often ended up being several thousand after the homing operation had completed.

PR #635, opened 2017-04-28
Added a notification for restart when the language is changed.

joacimb:
When the language settings are changed the platform needs to be restarted. With this PR these changes will notify the user about a platform restart.


winder:
Thanks, that's really nice!

Issue #637, opened 2017-05-01
StackOverflow when running GTK look and feel on Ubuntu


joacimb:
When running UGS platform on linux with the GTK look and feel I get this stacktrace and the Jog-panel isn't shown. The error can also be reproduced when running the WidgetPreview using the option -Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel

```
Exception in thread "main" java.lang.StackOverflowError
	at java.awt.KeyboardFocusManager.getCurrentKeyboardFocusManager(KeyboardFocusManager.java:216)
	at java.awt.Component.hasFocus(Component.java:8033)
	at java.awt.Component.isFocusOwner(Component.java:8046)
	at javax.swing.plaf.synth.SynthLookAndFeel.getComponentState(SynthLookAndFeel.java:217)
	at javax.swing.plaf.synth.SynthSpinnerUI.getContext(SynthSpinnerUI.java:277)
	at javax.swing.plaf.synth.SynthDefaultLookup.getDefault(SynthDefaultLookup.java:42)
	at sun.swing.DefaultLookup.get(DefaultLookup.java:125)
	at sun.swing.DefaultLookup.getBoolean(DefaultLookup.java:161)
	at javax.swing.plaf.basic.BasicSpinnerUI$Handler.stateChanged(BasicSpinnerUI.java:1053)
	at javax.swing.JSpinner.fireStateChanged(JSpinner.java:458)
	at javax.swing.JSpinner$ModelListener.stateChanged(JSpinner.java:386)
	at javax.swing.AbstractSpinnerModel.fireStateChanged(AbstractSpinnerModel.java:119)
	at javax.swing.SpinnerNumberModel.setStepSize(SpinnerNumberModel.java:302)
	at com.willwinder.universalgcodesender.uielements.jog.StepSizeSpinnerModel.getNextValue(StepSizeSpinnerModel.java:46)
	at javax.swing.plaf.basic.BasicSpinnerUI.updateEnabledState(BasicSpinnerUI.java:497)
	at javax.swing.plaf.basic.BasicSpinnerUI.updateEnabledState(BasicSpinnerUI.java:481)
	at javax.swing.plaf.basic.BasicSpinnerUI.access$200(BasicSpinnerUI.java:50)
	at javax.swing.plaf.basic.BasicSpinnerUI$Handler.stateChanged(BasicSpinnerUI.java:1057)
[...]

PR #638, opened 2017-05-01
#637 fixed stack trace in jog-panel when using GTK-theme

joacimb:
By calling getNextValue an event will be fired which will cause a call to the same method again. This will be repeated untill the stack is full. Added a test and rewrote the code so that it won't call getNextValue().

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/uielements/jog/StepSizeSpinnerModel.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/uielements/jog/StepSizeSpinnerModel.java
@@ -29,9 +29,6 @@ public class StepSizeSpinnerModel extends SpinnerNumberModel {
         super(1.0, MIN_VALUE, MAX_VALUE, 1.0);
     }
 
-
-
-
     @Override
     public Object getNextValue() {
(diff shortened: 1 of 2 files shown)


ugs-core/src/com/willwinder/universalgcodesender/uielements/jog/StepSizeSpinnerModel.java
```diff
@@ -43,8 +40,11 @@ public Object getNextValue() {
                 stepSize = stepSize / 10;
             }
         }
-        setStepSize(stepSize);
-        return super.getNextValue();
+
+        if( value >= MAX_VALUE) {
+            return 0;
```
winder: This should probably be `return MAX_VALUE`, to prevent cycling the step size back to zero. And below should be `return MIN_VALUE`

joacimb: Looked in to the api docs and i should actually return null. I checked with the old implementation and it did return null as well.
https://docs.oracle.com/javase/7/docs/api/javax/swing/SpinnerNumberModel.html#getNextValue()

PR #643, opened 2017-05-02
Macros as a plugin

breiler:
I've extracted the macros functionality as a plugin. This way it should be easy for users to switch the flavour of the macros functionality as they like.

The module is installed by default but could be removed using the plugins manager, which would remove everything from the macros window, the settings panel and menus:
screen shot 2017-05-02 at 04 57 36

Added a link to the macros editing panel in the menu as some users had problems finding it:
screen shot 2017-05-02 at 04 58 15

As it's an action its available to place in the toolbar (here shown as the icon furthest to the right). If pressed it will only show the macros window:
screen shot 2017-05-02 at 04 59 48

### DIFF
--- a/ugs-platform/ugs-platform-plugin-macros/pom.xml
+++ b/ugs-platform/ugs-platform-plugin-macros/pom.xml
@@ -32,36 +32,6 @@
         </dependency>
         <dependency>
             <groupId>org.netbeans.api</groupId>
-            <artifactId>org-openide-nodes</artifactId>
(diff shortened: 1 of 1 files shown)


winder:
Do the macro's show up in the keybinding menu with this approach? A lot of the things done in the `MacroService` were mainly to get that working. The menu looks really nice, I especially like the `Edit macros` option.

joacimb:
Damn, no you are right. I'll fix it so that it uses your service again.

PR #728, opened 2017-09-01
Use proper command names in UI instead of grbl command strings

pelrun:
I've changed the $H/$X/$C/$G button labels to their actual functions instead of the underlying GRBL commands.

It's only been done for English (I'm no polyglot) but it's ready for localisation.


winder:
Thanks, this should have been done a long time ago!

vMeph:
@winder @pelrun  , was wondering if the jog controler mode  coud be change the ( " ) to Inch or In?
dont know what would be better but anything will be better then ( " )
i think this is something that been ask to be changed in past afew times

thanks for the great work

PR #745, opened 2017-09-06
Force cast double to int to correct ClassCastException

oddish2211:
A fix for issue 744


winder:
The spinner should actually be a double for this one, I didn't configure the model properties correctly. I slapped this GUI together in the NetBeans swing builder and didn't change the number type from Int -> Double.

PR #748, opened 2017-09-06
Fixed missing retractHeight field in settings

oddish2211:
I believe this caused the failed build


winder:
Yeah, I accidentally pushed a partial commit. This should help until I can wrap it up.

PR #774, opened 2017-09-30
Update the default look and feel

derickfelix:
This change was made because it seems that only the ugs-platform has a look and feel according to the system, and this change makes the same thing with the ugs-core


pelrun:
I'd suggest removing all the formatting changes you (or your IDE) did to the files and only submit the actual code changes.

winder:
@derickfelix have you been using the "experimental" swing GUI? I don't think there's even a way to access that from a normal release. That file has been pretty much abandoned for 1.5yr.

If you're using the main swing UI and this updates the look and feel, it is a coincidence.  You should make this change in MainWindow.java instead.

PR #797, opened 2017-11-08
Basics for the API Server

acoll:
Very early version of the pendant server. Currently the pendant client is served at http://ugs-pendant.s3-website-us-east-1.amazonaws.com/ .

image


ugs-platform/ugs-platform-plugin-pendant/src/main/java/com/willwinder/ugs/platform/plugin/pendant/Bundle.properties
```diff
@@ -0,0 +1 @@
+OpenIDE-Module-Name=PendantModule
```
winder: Please add these fields, they are used in the "Plugins" section to describe the purpose:
```
OpenIDE-Module-Short-Description=
OpenIDE-Module-Long-Description=
OpenIDE-Module-Display-Category=UGS Plugin (or maybe a new category if this is a general purpose interface)
```

ugs-platform/ugs-platform-plugin-pendant/src/main/java/com/willwinder/ugs/platform/plugin/pendant/PendantTopComponent.java
```diff
@@ -0,0 +1,138 @@
...
+@TopComponent.OpenActionRegistration(
+        displayName = "#CTL_PendantAction",
+        preferredID = "PendantTopComponent"
+)
+@Messages({
+    "CTL_PendantAction=Pendant",
+    "CTL_PendantTopComponent=Pendant Window",
+    "HINT_PendantTopComponent=This is a Pendant window"
+})
+public final class PendantTopComponent extends TopComponent {
```
winder: There's probably no need to put this in a `TopComponent`. Take a look at `RunActionService` for an example of how to create a Netbeans `ServiceProvider`.

Services aren't initialized until they're used, so you would pair the `PendantServer` would need to be initialized with a `Lookup.getDefault().lookup(PendantServer.class);` call in an action like `PendantAction.java`.

ugs-platform/ugs-platform-plugin-pendant/src/main/java/com/willwinder/ugs/platform/plugin/pendant/PendantServer.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Called by the GUI Component to start the server.
+     *
+     * @throws Exception
+     */
+    public void start() throws Exception {
+        server = WebServers.createWebServer(port);
+
+//        server.add("/", new StaticFileHandler("./"));
```
winder: It looks like webbit might not support serving files in the classpath. I'll have to investigate what utilities exist for extracting files into the netbeans cache directories. I also manually extract resource files in other places but that's more of a plan B.

If you could include the static assets in a resource directory I'll do some experimenting.

ugs-platform/ugs-platform-plugin-pendant/src/main/java/com/willwinder/ugs/platform/plugin/pendant/PendantServer.java
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Sends the current UGS state down to the clients.
+     */
+    private String buildStateString(ControllerStatus cs) {
+        Map<String, Object> obj = new HashMap<>();
+
+        obj.put("availableCommands", PendantCommand.class.getEnumConstants());
+        obj.put("controllerStatus", cs);
+        obj.put("settings", backend.getSettings());
+        obj.put("isManualControlEnabled", isManualControlEnabled());
```
winder: No need for `isManualControlEnabled`, use `jogService.canJog()`

ugs-platform/ugs-platform-plugin-pendant/src/main/java/com/willwinder/ugs/platform/plugin/pendant/PendantServer.java
acoll: I was trying to use the EmbeddedResourceHandler first from
https://github.com/webbit/webbit/blob/master/src/main/java/org/webbitserver/handler/EmbeddedResourceHandler.java

but beyond that I couldn't even get an http response back from it.

winder:
@acoll this is a great start. Thanks for taking this on. Don't worry too much about those codacy errors, that service is much stricter than I am (as you can see my the global codacy report).

I left a few comments in the code with some feedback. It looks like the biggest challenge is going to be figuring out how to serve files with webbit. Webbit seems like a pretty small project so I'm wondering if something like jetty/netty would have better support for everything required.

ugs-platform/ugs-platform-plugin-pendant/src/main/java/com/willwinder/ugs/platform/plugin/pendant/PendantServer.java
winder: Oh perfect. Java can be a PITA with how you format your classpath string, I can help investigate that.

vMeph:
Is this feature already available?

acoll:
Hey @vMeph ,

Originally the intent here was to move the pendant logic into the new plugin format and out of core as well as pave the way for some future pendant improvements. I'd like to have a websocket connection between the pendant client and UGS instead of constant polling. After getting some of it working it occurred to me that maybe this was better suited as a general API or mechanism to build whatever you want to connect to UGS. One example of such is a pendant client.

I'm not sure of the right direction but am interested in thoughts from @winder and the community on which way to go with this.

Issue #812, opened 2017-11-20
Error while parsing response


dastultz:
Hello,

I just installed the latest nightly build. When I connect I get a series of errors parsing the state of GRBL. I'm on a Mac 10.13, Java 1.8, GRBL 1.1. Not sure how to describe the exact version of UGS. The screenshot says "Error while processing response >" in the error box. It repeats continuously, layering the error boxes.

<img width="756" alt="screen shot 2017-11-19 at 9 01 28 pm" src="https://user-images.githubusercontent.com/4334042/32999008-76dadc6c-cd6d-11e7-8b0a-f945ac0a34f9.png">

DeweyCheatem:
I have this same error with the Windows 32 build using Java 1.8 and GRBL1.1F

winder:
What version of UGS are you running?
Do the errors happen right away or do they start after using the program for some time?

DeweyCheatem:
I've tried both of the latest UGSPlatform nightly builds.
This error happens right away on connect. I'm not able to do anything (even disconnect). I'm usually forced to unplug the USB cable from the computer.

dastultz:
I don't know who you are asking. I don't know how to tell what version of UGS I am using. Please advise. The errors start just ... oh, Mr Howe just replied. Same thing, they start just after connection and all I can do is exit the program.

dastultz:
This regression was introduced by #739. Code assumes if `getFeedSpeed()` is not null, `getSpindleSpeed()` is also not null. I do not have spindle control on my machine so it fails with NPE. It's easy enough to fix. Is it desirable to have 2 if/else blocks, one for feed/speed the other for spindle?

PR #821, opened 2017-11-26
Fix NPE in machine state feed/spindle speed display.

dastultz:
There are four paths through this code. I tested 2 of them. Not sure what conditions for which `feedSpeed` is null. I don't have spindle control. So I tested `feedSpeed` not null and `spindleSpeed` null.

issue #812


winder:
Thanks, I was just starting to look at this.

feed and spindle speed were both added at the same time in GRBL 1.1, so they should either both be available or missing. I thought the old code was correct but apparently not.

PR #829, opened 2017-11-30
DRO Panel visual improvements

dastultz:
Earlier changes I made to the DRO don't play well with various Look and Feel "themes".
Derive color of sub panels from parent container.
Preserve default text color.
Tweak size of machine position font to improve rendering.


winder:
I think the DRO panel has always been broken with the dark theme, so this is fixing a day zero bug - nice!

PR #845, opened 2017-12-09
Dro

dastultz:
Hi Will, 

Rework of the DRO panel.
Moved latest comment to send status panel.
Added a couple fonts, noted source in comment.

I don't know how you manage integration of the translation system. I removed a few strings no longer used from the code but not all from the bundle, and added some, perhaps with the wrong group (maybe should have used "swing" group for some). Perhaps you need to reconcile all this with the translation system.

I don't have any pins to test out the alarm pin state support. Oddly GRBL reports the probe pin is raised so that seemed enough to demonstrate it was working.

All in all, there's a lot here and if you can you should test it before merging. In particular, try out resizing the panel.


ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/SendStatusPanel.java
```diff
@@ -196,6 +204,19 @@ private void initComponents() {
...
+        add(latestCommentValueLabel, "growx, span 2, wrap");
+
+        latestCommentValueLabel.setOpaque(false);
+        latestCommentValueLabel.setWrapStyleWord(true);
+        latestCommentValueLabel.setLineWrap(true);
+
+        addComponentListener(new ComponentAdapter() {
+            @Override
+            public void componentResized(ComponentEvent event) {
+               latestCommentValueLabel.setMaximumSize(new Dimension(SendStatusPanel.this.getWidth() - 10, 30));
```
winder: Is this to make sure the comment label is padded? There are options to do this sort of thing with MigLayout I can lookup.

ugs-core/src/resources/MessagesBundle_en_US.properties
```diff
@@ -32,9 +32,7 @@ mainWindow.helpKeyPlusMinus = In-/Descrease Steps\: KeyPad +/-
 mainWindow.helpKeyDivMul = Change Steps\: KeyPad DIVIDE/MULTIPLY
 mainWindow.helpKeyZero = Reset Zero\: KeyPad Insert; Numpad\: 0
 mainWindow.status.alarm = Alarm
-mainWindow.status.hold = Hold
-mainWindow.status.queue = Queue
-mainWindow.status.run = Run
```
winder: Pretty much all languages have these fields already, do they need to be removed?

ugs-core/src/com/willwinder/universalgcodesender/uielements/helpers/FontManager.java
```diff
@@ -0,0 +1,148 @@
...
+import javax.swing.JComponent;
+import javax.swing.JLabel;
+import java.awt.Font;
+import java.awt.GraphicsEnvironment;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class FontManager {
```
winder: Is this used yet?

ugs-core/src/com/willwinder/universalgcodesender/uielements/helpers/SteppedSizeManager.java
```diff
@@ -0,0 +1,79 @@
...
+import java.awt.event.ComponentAdapter;
+import java.awt.event.ComponentEvent;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Track size of a component and fire event when nominal size changes.
+ */
+public class SteppedSizeManager {
```
winder: Curious to see where you're going with this, not used yet?

ugs-core/src/com/willwinder/universalgcodesender/uielements/helpers/FontManager.java
winder: This is pretty specific to the DRO panel right now, were you thinking it would evolve into a general purpose thing? Could it be renamed to MachineStatusFontManager for now?

ugs-core/src/com/willwinder/universalgcodesender/uielements/helpers/SteppedSizeManager.java
winder: Intrigued by this. Could the sizes be used for things like buttons and images as well? Or is it mainly for labels?

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
```diff
@@ -121,116 +120,156 @@ public MachineStatusPanel(BackendAPI backend) {
...
+        fontManager.addWorkCoordinateLabel(work);
+        fontManager.addMachineCoordinateLabel(machine);
+
+        add(axisPanel,"growx, span 2");
+
+        resetPanel.addClickListener(() -> resetCoordinateButton(axis) );
+        axisResetControls.add(resetPanel);
+    }
+
+    private Timer createTimer() {
```
winder: The `ControllerListener` is the deprecated way to get these events, and `ControllerStatusListener` is the new version to get it. Did you switch to this method for performance reasons?

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerListenerAdapter.java
```diff
@@ -0,0 +1,38 @@
+package com.willwinder.universalgcodesender.listeners;
+
+import com.willwinder.universalgcodesender.model.Position;
+import com.willwinder.universalgcodesender.model.UGSEvent;
+import com.willwinder.universalgcodesender.types.GcodeCommand;
+
+public class ControllerListenerAdapter implements ControllerListener {
```
winder: I see what you did now. `ControllerListener` is sort of deprecated because I don't like having all the unused methods either. You can use `ControllerStateListener` to get this data instead and it utilizes the same `UGSEvent` callback.

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
```diff
@@ -121,116 +120,156 @@ public MachineStatusPanel(BackendAPI backend) {
...
+                activeStateValueLabel.setText(OFFLINE);
+                if (backend.getSettings().isDisplayStateColor()) {
+                    activeStatePanel.setBackground(Color.BLACK);
+                }
+            }
+            GcodeState state = backend.getGcodeState();
+            if (state == null) {
+                gStatesLabel.setText("--");
+            } else {
+                StringBuilder sb = new StringBuilder();
```
winder: There's a `String.join` method in Java8 so you can do something like:
```
              String.join(" ",
                      state.currentMotionMode.toString(),
                      state.units.toString(),
                      state.feedMode.toString());
```

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
```diff
@@ -381,41 +382,35 @@ public void statusStringListener(ControllerStatus status) {
...
-                foreground = Color.BLACK;
-            } else if (state.equals(Localization.getString("mainWindow.status.queue"))) {
-                background = new Color(220, 200, 0);
-                foreground = Color.BLACK;
-            } else if (state.equals(Localization.getString("mainWindow.status.run"))) {
-                background = new Color(0, 128, 0);
-                foreground = Color.WHITE;
-            } else {
-                this.activeStatePanel.setGradientEnabled(false);
+            Color background = ThemeColors.GREY;
```
winder: While you're fixing this could you update the `activeStateValueLabel` with the localized string?

winder:
Just gave this a once over and it looks really good. I've got some holiday projects to work on tomorrow so I'll get some testing in to really see it in action.

screen shot 2017-12-09 at 12 36 30 am

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/SendStatusPanel.java
dastultz: I spent quite a lot of time trying to get the last comment text to do what I wanted it to. When it was on the DRO I could get the comment text to do what I wanted, then the other components would break. I wanted the text to wrap and it proved rather challenging. If it can be done straight in MigLayout I'm happy to see that solution, I just don't think I can figure it out myself!

ugs-core/src/resources/MessagesBundle_en_US.properties
dastultz: Sorry, I don't think I understand what you are saying. I'm not sure what your approach to managing translations is. Should I have not removed these?

ugs-core/src/com/willwinder/universalgcodesender/uielements/helpers/FontManager.java
dastultz: I expect to use this in the jog panel as well. I first started with this as a Singleton - no need to load the fonts more than once. Then I started adding listeners and such. I don't have a problem calling it MachineStatusFontManager and then just copying it for jog panel use. I'll see if I can figure out something better.

ugs-core/src/com/willwinder/universalgcodesender/uielements/helpers/SteppedSizeManager.java
dastultz: Because of the apparent need to pre-generate fonts at specific sizes, it's meant for labels. (That's why it's "stepped" rather than "continuous".) But sure, it can be used for anything that can be resized.

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
dastultz: Assuming you commented the wrong code. I'll respond to the next comment on this issue.

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerListenerAdapter.java
dastultz: An Adapter is common pattern to use when only one or a few of the interface methods are interesting to implement, then you don't have all those blank methods. This was purely to reduce the number of lines of code. I'll look into ControllerStateListener. I was afraid of breaking functionality here.

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
dastultz: Oh cool, I knew such a thing was in Kotlin. My J8 experience is light...

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
dastultz: Oh snap, missed that one, thanks. 
Is this what you want me to fix?
```this.activeStateValueLabel.setText( status.getState() );```
What's the approach you use to change GRBL literals to translated values? Something like this?
```Localization.getString("mainWindow.status." + status.getState().toLowerCase()))```. I'm inclined to grab i18n values up front for performance reasons, probably use a `Map` of GRBL status to translated value.

winder: I would swap in the localizations right into these conditions. It looks like Jog might be missing, you can add that right into the english languange bundle. The online service I use for localization will synchronize terms later so no need to add it to every file.

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/SendStatusPanel.java
winder: Ugh, yeah Swing text wrapping is awful. I took a stab at it once with `NarrowOptionPane`, I think the recommended approach is to embed HTML (seriously).

ugs-core/src/resources/MessagesBundle_en_US.properties
dastultz: I think this ties in to another comment, I think I get it, put them back in, then map GRBL's status to these strings.

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/SendStatusPanel.java
dastultz: Marking as "won't fix" for tracking my tasks. If this is important to change, please give guidance.

ugs-core/src/resources/MessagesBundle_en_US.properties
dastultz: Marking as done for tracking my tasks.

ugs-core/src/com/willwinder/universalgcodesender/uielements/helpers/FontManager.java
dastultz: Marking as done for tracking my tasks.

ugs-core/src/com/willwinder/universalgcodesender/uielements/helpers/SteppedSizeManager.java
dastultz: Marking as "nothing to do" for tracking my tasks.

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
dastultz: Marking as done for tracking my tasks.

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerListenerAdapter.java
dastultz: Marking as done for tracking my tasks.

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
dastultz: Marking as done for tracking my tasks.

ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/MachineStatusPanel.java
dastultz: Marking as done for tracking my tasks.

dastultz:
I think I have fixed what needs to be fixed. Let me know.

vMeph:
At first look my DRO dont shows feed,speed G"s, and it seems so big
Im using dracula theme

dastultz:
On Tue, Dec 12, 2017 at 3:00 AM, Meph1 <notifications@github.com> wrote:

> At first look my DRO dont shows feed,speed G"s, and it seems so big
> Im using dracula theme
>

The G's don't show until you connect. Are you saying feed and speed is slow
to show or they don't show? If they show, when do they show?

The DRO is scalable in 3 steps, small, medium, and large. I don't think the
small is much bigger, if any than previous version.

/Daryl

vMeph:
this is how i see now it fells so big

wish there was ways to shrink it down the panel

dastultz:
Do you want it narrower or shorter? If you close the common actions / state panel so you have only the DRO in that column, then drag the split pane (between the DRO and visualizer) to the left you can make the DRO short and narrow. If you then put a taller panel above or below the DRO, you can make the DRO wider but it won't get taller. Play around with the panel dividers and the panels that are above and below the DRO to see how it changes size. Basically it tries to be as big as it can without getting cropped.

Here is the old and new side by side with the new at its smallest / narrowest. The old pictured here does not have the enabled pin state (ALARM) or the G-state values (--). So the common elements make it only slightly taller than the old. A number of people have said larger is better due to vision challenges.  I'd be willing to tweak this and make it no taller than the old if some other folks also object.

old_new_dro

vMeph:
i had before rows but if you put 3 rows then will hide the DRO

dastultz:
OK, I was under the impression that people wanted it bigger. I'll tweak the smallest size to be no bigger than the original.

vMeph:
i dont know what others think :(
was just saying my personal opinion

PR #854, opened 2017-12-13
split pendant gCode on semicolon to allow multiple commands in a single request

esasiela:
this matches the behavior for how macros handle embedded semicolon

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/MacroHelper.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/MacroHelper.java
@@ -27,6 +27,9 @@ public class MacroHelper {
     /**
      * Process and send a custom gcode string.
      * 
+     * Not safe to run in the AWT Event Dispatching thread, be sure to use
+     * EventQueue.invokeLater() if running in response to a GUI event.
+     * 
      * Interactive substitutions can be made with special characters:
      *   %machine_x% - The machine X location
      *   %machine_y% - The machine Y location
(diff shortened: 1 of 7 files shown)


winder:
I've never really used the pendant, does it allow you to use your macros from the web interface?

If so take a look at `MacroHelper` your change would basically be the same as this but use `MacroHelper.executeCustomGcode(gCode, mainWindow)`

esasiela:
yeah, i looked at that first.  the issue is that MacroHelper.executeCustomGcode() uses EventQueue.invokeLater() because it runs in the GUI event dispatching thread from the action listener for the macro buttons.  The pendant UI is in its own jetty thread, so can safely avoid using invokeLater().  if you think it doesnt hurt anything to have an extra invokeLater(), then I can switch the PendantUI code to re-use executeCustomGcode.

esasiela:
and to answer your first question about the pendant, no it does not expose the macros via the pendant.

winder:
I think having the extra invokeLater is fine to make sure macros are handled in a consistent way

esasiela:
ha. switching it from split(";") to MacroHelper.executeCustomCode() made the unit test fail, so I'll have to dig a bit deeper to see what the test is doing/expecting.

winder:
Oops, maybe lifting that invoke later out is an option too? Looks like executeCustomGcode is only called in 2 places (and a test)

esasiela:
yeah, i would feel better moving the invokeLater's outside the macrohelper code, then each caller can decide for itself what the right threading policy is. will require small updates to MacroService, MacroActionPanel, MacroPanel, MacroHelperTest, plus PendantUI and PendantUITest.  Along the way, i'll move the exception handling outside of MacroHelper too, to allow each invoker to decide for itself how best to handle errors.  it might take me a little bit to figure out the unit test mechanism, but the other parts are not too bad.

winder:
This is really old code, I may not have known what I was doing back then. I really appreciate any cleanup you can do here.

I'm almost certain tests are failing because invoke later causes the method to happen in another thread after the assertions are checked. I've had this problem with other tests as well and have some pretty dodgy workarounds for it.

esasiela:
i updated the code to reuse the split/loop logic from the macro code.  also moved event thread and exception handling out of the split/loop code to allow invokers to handle as appropriate.  also updated the impacted unit tests.

winder:
Looks good. Could you switch the tabs to spaces? Ready to merge after that.

Thanks for the contribution!

esasiela:
should be all set with spaces, unless i missed one.  is there an auto checker for that as part of github? i didn't want to reformat entire files so hopefully i got all the impacted lines.

Issue #939, opened 2018-03-09
Jog controller design


breiler:
With some help from @dastultz, we've started building a new jog controller plugin.

The intention is to make it more scalable than the previous component and have a consistent look with the DRO. It's still a bit of a work in progress but I would like your comments and feedback.

The normal view:
![screenshot 2018-03-09 16 31 37](https://user-images.githubusercontent.com/8962024/37215120-5ab529a2-23b7-11e8-90de-cd45b2c869ef.png)

The bit more compact view:
![screenshot 2018-03-09 16 33 38](https://user-images.githubusercontent.com/8962024/37215223-a60a82c6-23b7-11e8-918a-6ba8c2cb44dc.png)

vMeph:
Looks awsome, if you keep button press will it keep moving until you release button also?

winder:
Looks great!

This next comment is a bit outside the scope of this redesign, but I'm curious if you saw this. There were some ideas in #779 for a way to make jogging more responsive to keyboard commands (i.e. how it works in the classic GUI). My original idea was to have a button that makes this panel pop up in a modal way, but maybe there is another solution?

The problem is that keyboard commands are sometimes swallowed (I think by the visualizer, and the text editor). When this happens keyboard shortcuts like jogging don't always work. Also simple single-character commands don't work well with the NetBeans keymappings.

breiler:
@vMeph I first implemented it like described here: https://github.com/gnea/grbl/wiki/Grbl-v1.1-Jogging#joystick-implementation
It would jog as long as you pressed the jog-button, on release a jog-cancel command was sent. It worked pretty well but had the side effect that it would be very hard jogging to an exact position. So I scrapped that, it's now working as before. I was thinking of adding a button-pressed-event where continuous jogging is activated after the button  is being pressed for ~500ms.

@winder yes I saw it, this could be a good next step. I saw that sourcerabbit have button for giving the panel focus: https://raw.githubusercontent.com/nsiatras/sourcerabbit-gcode-sender/master/Images/ControlForm.png

jahnj0584:
Is this the new colour scheme UGS is heading towards? Could we have options
for the standard netbean tan + black? I love black and cyan but i also want
my control to be as simple as possible.

breiler:
That's a good input. It would actually make the coding a lot more simple if we used standard buttons and we would use more of the panel for the buttons (less borders). 

Just for comparison, this is what it would look like in standard colors (it still needs some adjustments):
[...]

PR #947, opened 2018-03-15
Jog module

breiler:
Se the comments in #939


ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/JogTopComponent.java
```diff
@@ -0,0 +1,100 @@
...
+        path = JogTopComponent.WINOW_PATH)
+@TopComponent.OpenActionRegistration(
+        displayName = "Jog Controller",
+        preferredID = "JogTopComponent"
+)
+public final class JogTopComponent extends TopComponent implements UGSEventListener {
+
+    public static final String WINOW_PATH = LocalizingService.MENU_WINDOW_PLUGIN;
+    public static final String CATEGORY = LocalizingService.CATEGORY_WINDOW;
+    public static final String ACTION_ID = "com.willwinder.ugs.nbp.jog.JogTopComponent";
```
winder: I made a new pattern for localizing so that plugins don't need to modify `LocalizingService`, this case may be a little different though because there are two jog modules (and it doesn't handle the actions), you would drop something like this in right here:
```
    @OnStart
    public static class Localizer extends TopComponentLocalizer {
      public Localizer() {
        super(VisualizerCategory, VisualizerActionId, VisualizerTitle);
      }
    }
```

winder: Something like this for registering action menu's might work as well.

PR #949, opened 2018-03-15
Add popup to jog menu for quick access to separate Z axis setting.

winder:
@breiler @dastultz a right-click popup menu is used for the console settings, do you think it would be good here as well?

screen shot 2018-03-15 at 3 29 51 pm


dastultz:
Is that a right-click action? If so, is it expected users will just "discover" it?

winder:
Yes, it is a right-click popup menu. You're right there's no way they would know about it unless they happened to be right clicking things.

breiler:
Using separate step sizes is for intermediate users and I wouldn't mind having these kinds of "shortcuts" in the application. 

The only thing I have against it is that you need to hit the panel with the mouse click for this to work.

My vote is, go for it =)

Issue #951, opened 2018-03-16
New Jog Controller Error 15 when holding down mouse button (Build 966)


kenfritz:
Today I tried using the new Jog Controller but got an "Error 15: Jog Target Exceeds Machine Travel" when clicking on any of the jog arrows and holding down the left mouse button more than 1 second. If I single click the mouse button I don't get an error. This error is not a single error but starts a chain reaction of the error repeating hundreds of time and I must Kill UGS process to get out of this problem.

I attached a screen shot showing the error.
Build 966 (UGS Plateform)
![ugs build 966 error using new jog controller](https://user-images.githubusercontent.com/11965304/37500568-6f499086-2887-11e8-8756-cda45a1e54f0.jpg)

breiler:
Argh, sorry about that!

I haven't tested it with soft limits, and now that I see the error I feel stupid! 😓 

When the button is pressed and hold it will continuously queue jog commands with a quite big step length so that the machine is kept busy with processing commands. According to the GRBL documentation these steps should be kept small, but I got in to trouble when I tried that because it ended up streaming at some point which lead to that the jog cancel didn't work (maybe because the send buffer was filled).

I'll try to fix this by checking the send queue before creating a new jog command, making sure just one or two commands are queued. Then decrease the step length, maybe base it on the current units and feed rate.

breiler:
Fixed some of the problems:
* Shortened the jog size so that the soft limit isn't triggered directly
* It will now cancel the jog command if there's an error (instead of leaving the jog-loop to continuing) 

One problem still remains. There is no way to find out how many jog commands that are queued which means  there is no way of limiting the number of commands that are being sent.

breiler:
@kenfritz this should be available in the nightly build. If you get a chance to test it please let me know if it is working better for you.

kenfritz:
I have been testing today Bld# 970 jogging around and it looks fixed. I jogged long distances holding the mouse button and got no errors. I loved the diagonal arrows the best. I really like the changes you are making to the jog control feature. Great job.

PR #954, opened 2018-03-17
Reworked the plugin so that it doesn't end up in a endless loop 

breiler:
There was some problems when doing a continuous jog if soft limits where enabled (issue #951)
* Changed the jog distance to 5mm or 0.2" if  doing a continuous jog
* Moved the long press logic to its own mouse listener class
* Moved the action logic from the panel to the top component
* Listens to processed commands and cancels any jog commands if there is an error


winder:
I think that the GrblController has all the settings cached, so it might be possible to extend the existing jog logic there to validate soft limits before sending the command.

breiler:
I'm sorry I don't follow. How do you mean validating the soft limits before sending?

Issue #759, opened 2017-09-11
Bad state after an error is detected in $C mode


jahnj0584:
Set grbl to $c. Send a file that failed with error 3 or whatever it is were already trying to solve. 
Soft reset still won't let me hit send as it's "currently streaming" even though file is done. If you do $x or soft reset it does nothing. Disconnecting then connecting again will let you back but the jog controls are disabled. You have to restart the program to remedy this. 
![20170910_214144](https://user-images.githubusercontent.com/17396492/30254571-2579afd4-9671-11e7-8ed1-9b872e48c9e5.jpg)

jahnj0584:
yep, this is definitely just an issue if you send a file as $c and it has an error 33 or 20 (or probably any).
The file never finishes because the error so the GUI always thinks its still streaming a file.

winder:
I noticed this the other day, grbl is in `Check` mode which I don't handle properly when the stream is paused due to an error.

jahnj0584:
I recommend you make it cancel the stream after an alarm pop up. The
operator the. Fixes it and parses it again to see. If there's still a
mistake it'll be shown. Rinse and repeat until they give up or it parsed
properly 😅

jahnj0584:
Can confirm this happens on the Sep10 version

vMeph:
this bug always been present for a long time.
only solution is close UGS and restart program
would love to see this fix also...

breiler:
I've done some digging.

I added the communicator to the diagnostics panel to get some more information. And reproduce the problem:

1. Home my machine
2. Reset to Zero
3. Open circle.gcode that comes with the test files for UGS
4. Use "Check Mode"
5. Run the file and get the following error and close it (on my test rig this is a correct error because the soft limits will be hit if I try to go any higher on Z than 0):
![screenshot 2018-03-19 19 43 40](https://user-images.githubusercontent.com/8962024/37615572-eea50422-2bad-11e8-82b7-61315cdb6e62.png)
6. Click "Soft reset" as the message in the console tells me
7. Click "Unlock" because we are still in an error state, and nothing happens. The console doesn't show that any commands are sent.
8. Refresh the current state in my custom UGS Diagnostics and can see that the communicator is still paused even though I've done a soft reset.
![screenshot 2018-03-19 19 47 45](https://user-images.githubusercontent.com/8962024/37615872-9135b98e-2bae-11e8-931f-869c63b0e4c4.png)

[...]

PR #967, opened 2018-03-23
Check mode

breiler:
Fixes for #759


ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
```diff
@@ -722,11 +723,26 @@ public void commandSent(GcodeCommand command) {
         dispatchCommandSent(command);
     }
 
+    @Override
+    public void communicatorPaused() {
+        dispatchConsoleMessage(MessageType.INFO, "**** The communicator has been paused ****\n");
+        try {
+            // FIXME the dispatchStateChange has logic for resuming send operations, this should have it's own handler.
```
winder: This could use a comment that it is specifically circumventing the `setCurrentState`/`handleAllStateChangeEvents` in case of an error (also see the other comment about renaming this callback).

ugs-core/src/com/willwinder/universalgcodesender/GrblController.java
```diff
@@ -477,6 +495,13 @@ public void toggleCheckMode() throws Exception {
             if (!"".equals(gcode)) {
                 GcodeCommand command = createCommand(gcode);
                 this.sendCommandImmediately(command);
+
+                if (getControlState() == COMM_CHECK) {
```
winder: This if/else should probably be in `handleStatusString` (i.e. in case a soft reset or external button press changes the state out of CHECK mode)

ugs-core/src/com/willwinder/universalgcodesender/GrblController.java
```diff
@@ -640,8 +665,11 @@ private void handleStatusString(final String string) {
             if (attemptsRemaining > 0 && lastLocation != null) {
                 attemptsRemaining--;
                 // If the machine goes into idle, we no longer need to cancel.
-                if (this.controllerStatus.getState().equals("Idle")) {
+                if (this.controllerStatus.getState().equals("Idle") || this.controllerStatus.getState().equalsIgnoreCase("Check")) {
                     isCanceling = false;
+
+                    // Make sure the GUI gets updated
+                    this.dispatchStateChange(getControlState());
```
winder: Do you remember which edge case this happens in?

ugs-core/src/com/willwinder/universalgcodesender/listeners/SerialCommunicatorListener.java
```diff
@@ -27,15 +22,33 @@ This file is part of Universal Gcode Sender (UGS).
...
  *
  * @author wwinder
  */
 public interface SerialCommunicatorListener {
-    //void capabilitiesListener(CommUtils.Capabilities capability);
     void rawResponseListener(String response);
-    
+
+    /**
+     * This method will be called when a command has
```
winder: Missing the "been sent."

ugs-core/src/com/willwinder/universalgcodesender/model/GUIBackend.java
```diff
@@ -76,7 +76,6 @@ This file is part of Universal Gcode Sender (UGS).
     private String lastComment;
     private String activeState;
     private long estimatedSendDuration = -1L;
-    private boolean sendingFile = false;
```
winder: 👍

ugs-core/src/com/willwinder/universalgcodesender/model/GUIBackend.java
```diff
@@ -586,7 +584,7 @@ public void pauseResume() throws Exception {
             switch(getControlState()) {
                 case COMM_IDLE:
                 default:
-                    if (!sendingFile) {
+                    if (!this.controller.isStreaming()) {
```
winder: Use `isSendingFile()` for the null check.

ugs-core/src/com/willwinder/universalgcodesender/model/GUIBackend.java
```diff
@@ -720,40 +716,38 @@ public void commandSent(GcodeCommand command) {
...
     @Override
     public void commandComplete(GcodeCommand command) {
-        if (command.isError()) {
-            if (this.sendingFile && !this.isPaused()) {
-                try {
-                    this.pauseResume();
-                } catch (Exception e) {
-                    GUIHelpers.displayErrorDialog(e.getLocalizedMessage());
-                }
+        if (command.isError() && this.controller.isStreaming() && !this.isPaused()) {
```
winder: Use `isSendingFile()` instead of `controller.isStreaming()` for the null check.

ugs-core/src/com/willwinder/universalgcodesender/model/GUIBackend.java
```diff
@@ -720,40 +716,38 @@ public void commandSent(GcodeCommand command) {
...
+            try {
+                this.pauseResume();
+            } catch (Exception e) {
+                GUIHelpers.displayErrorDialog(e.getLocalizedMessage());
+            }
 
-                /*
```
winder: This commented out code can be deleted, it caused a ton of problems there's no way we'll want to re-enable it.

ugs-core/test/com/willwinder/universalgcodesender/GrblCommunicatorTest.java
```diff
@@ -356,7 +356,7 @@ public void testCancelSend() {
         instance.streamCommands();
         instance.cancelSend();
         // Verify that there are several active commands.
-        expectedBool = true;
+        expectedBool = false;
```
winder: Please update the test comment here, something like "Make sure cancelSend clears out the commands."

ugs-core/src/com/willwinder/universalgcodesender/BufferedCommunicator.java
```diff
@@ -195,20 +197,24 @@ else if (!this.commandBuffer.isEmpty()) {
     synchronized public void streamCommands() {
         // If there are no commands to send, exit.
         if (this.getNextCommand() == null) {
+            logger.log(Level.INFO, "There are no more commands to stream");
             return;
         }
         
         // If streaming is paused, exit.
-        if (this.sendPaused) {
+        if (isPaused()) {
```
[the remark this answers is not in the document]
breiler: Right, I used this to debug why commands wasn't processed. But it's now visible in the diagnostics so it's not needed anymore.

ugs-core/src/com/willwinder/universalgcodesender/BufferedCommunicator.java
```diff
@@ -130,14 +130,16 @@ public String activeCommandSummary() {
     @Override
     public boolean areActiveCommands() {
         return (this.activeCommandList.size() > 0)
-                || (this.commandStream != null && this.commandStream.getNumRowsRemaining() > 0);
+                || (this.commandStream != null && this.commandStream.getNumRowsRemaining() > 0)
+                || nextCommand != null;
```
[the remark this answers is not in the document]
breiler: Lol, right you are! =)

ugs-core/src/com/willwinder/universalgcodesender/GrblController.java
breiler: I think this was when pressing the cancelSend-button when paused it will normally change from states "Hold" to "Idle" and this will automatically dispatch a state change event. But when using "Check" we are simulating the "Hold/Pause" and no state change event will be triggered and we need to do that manually.

ugs-core/src/com/willwinder/universalgcodesender/GrblController.java
breiler: This took some work because the GRBL issues a soft reset when exiting Check-mode clearing the previous state. So there is nothing to compare with when trying to reset the single step mode. I solved this by not clearing the old status if we are in Check-mode.

PR #972, opened 2018-03-26
Handle streaming errors in the communicatorPausedOnError callback.

winder:
Handle streaming errors in the `communicatorPausedOnError` callback. Now that the controller is notified directly about communicator errors, the handling can be simplified quite a bit. It seems to work properly.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
@@ -735,6 +735,8 @@ public abstract class AbstractController implements SerialCommunicatorListener, 
             }
             else {
                 this.pauseStreaming();
+                // In check mode there is no state transition, so we need to manually make the notification.
+                this.dispatchStateChange(COMM_SENDING_PAUSED);
             }
         } catch (Exception ignored) {
             logger.log(Level.SEVERE, "Couldn't set the state to paused.");
--- a/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
@@ -804,11 +804,6 @@ public abstract class AbstractController implements SerialCommunicatorListener, 
      * that the stream is complete once the last command has finished.
      */
     public void commandComplete(String response) throws UnexpectedCommand {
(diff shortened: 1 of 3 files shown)


ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
```diff
@@ -886,10 +888,6 @@ protected void dispatchConsoleMessage(MessageType type, String message) {
     }
     
     protected void dispatchStateChange(ControlState state) {
-        if (this.currentState == COMM_SENDING_PAUSED && state != this.currentState
-                && this.comm.isPaused() && !this.isStreaming()) {
-            this.comm.resumeSend();
-        }
```
winder: This block was only relevant when `dispatchStateChange` is called from `communicatorPausedOnError`, so I moved it there.

winder:
There still seems to be an issue with the stop action not always getting enabled if you send a file that starts with an error repeatedly.

ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
```diff
@@ -729,11 +729,13 @@ public void commandSent(GcodeCommand command) {
...
-            // We can not use setCurrentState-method because in some cases the state change will not be dispatched
-            // if handleAllStateChangeEvents returns true.
-            dispatchStateChange(COMM_SENDING_PAUSED);
-            this.currentState = COMM_SENDING_PAUSED;
+            // Synchronize the controller <> communicator state.
+            if (!this.isStreaming()) {
+                this.comm.resumeSend();
+            }
+            else {
+                this.pauseStreaming();
```
breiler: This will not dispatch a state change paused event if handleAllStateChangeEvents returns true. This would mean if we were running Check-mode and an error occurred the GUIBackend wouldn't get the paused event and activate the stop button. I must have missed a test for this.

To reproduce the problem run this in check mode and try to stop the file after the first error:
```gcode
G00 X22.375 Y54.9375
M3 S80
G01 X21.5625 Y54.9375
M5
G00 X21.125 Y54.9375
M3 S80
G01 X20.875 Y54.9375
M5
```

To make this work this must be added as well or if you come up with something better:
`dispatchStateChange(COMM_SENDING_PAUSED);`

PR #973, opened 2018-03-27
Icons dark theme

breiler:
I've created icons for the dark theme. I'm not super pleased but they are at least usable now instead of before when they were black.

screenshot 2018-03-27 09 53 47


dastultz:
It looks pretty good to me. I'd like to see the "connect" icon without a border. It's not needed since we're on a dark background.

Also, you need a dark-theme color scheme for the visualizer, it helps a lot. :-)

breiler:
It would be nice to have visualizer color presets but I think it could wait.

Issue #738, opened 2017-09-04
More advanced probing plugin


winder:
Most of the features in this video should be straight forward to implement: https://youtu.be/pQGnV85Cu_A?t=191

I'll probably go ahead and make a bunch of screens that go over the visualizer for these:
- [ ] Calibration (a.k.a. probe / end mill size measurement)
- [ ] Corner probing (inside / outside)
- [ ] Angle measuring.
- [ ] Single surface probing (X / Y / Z)
- [ ] Selectable work coordinate system for X & Y probing (G43.1 doesn't work or make sense for these).

vMeph:
That looks so good and promissing!!!!

jahnj0584:
Wicked. I guess we're heading towards a tabbed controller due to all the
new features? Could you make it in "workspaces" so we can have eg. DRO,
visualizer, file control, overrides in one then another workspace for
macros, probing, settings, etc? All user controlled of course ;)

aforww:
Fantastic

winder:
I updated the existing Z probe button tonight. Since the X/Y probing didn't work I replaced that setting with a retract field. There's also a surprise between the initial probe and the final retract that I won't spoil. :)

<img width="203" alt="screen shot 2017-09-06 at 8 34 47 pm" src="https://user-images.githubusercontent.com/125509/30140470-dd89f750-9342-11e7-8ffc-f0ef582746f6.png">

groberts22:
Here are a couple of thoughts on the probe function.

The current version of the probe function does not display the current units, which can be very confusing.
The two most common uses of the probe (at least for me) are
1. Resetting the z offset after a tool change, and 
2. setting the xyz zero values for a new workpiece.
For both of these, I use a touchplate with known xyz thickness values that sits directly on the wasteboard of the machine or fits over the corner of the workpiece.  I think this is what many/most people do when they get sick of using a piece of thin paper and a calculator.
What would be most helpful (to me) is to have a helper that would allow me to position the tool at a reasonable distance from the surface(s) of the touchplate and then have the machine automatically set the relevant zero values for me.
The display could look similar to the current one, with an option to set a single axis value only (typically z) or all axes at once, and an option to choose the target WCS offset (54-59). [It might also be necessary to allow separate thickness and distance/direction settings for x,y, and z.]
The function could then use the G38.2 and G10L20 commands.
Thanks for your consideration.

jahnj0584:
[...]

PR #979, opened 2018-03-31
Simplify the probe module.

winder:
Removed a lot of overcomplicated state machine logic, this should make it much easier to add the remaining probe routines in #738


tkms002:
how do I get the latest PROBE MODULE?

PR #985, opened 2018-04-03
Controller state

breiler:
Added a ControllerState-enum with all states that should be supported by the GUI. These states will live alongside the controller state string until all usages have been replaced which will be done gradually to make sure no serious regressions is introduced. The old state string has been deprecated.

The MachineStatus-panel and isolated parts of GrblController are now using the new enum.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerStatus.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerStatus.java
@@ -34,26 +34,29 @@ public class ControllerStatus {
     private final OverridePercents overrides;
     private final EnabledPins pins;
     private final AccessoryStates accessoryStates;
+    private final ControllerState state;
 
     /**
      * Baseline constructor. This data should always be present. Represents the
      * controller status.
      * @param stateString controller state, i.e. idle/hold/running
+     * @param state controller state, i.e. {@link ControllerState#IDLE}/{@link ControllerState#HOLD}/{@link ControllerState#RUN}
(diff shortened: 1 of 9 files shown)


ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerStatus.java
```diff
@@ -34,25 +34,28 @@ This file is part of Universal Gcode Sender (UGS).
...
     }
 
     /**
      * Additional parameters
      */
-    public ControllerStatus(String state, Position machineCoord,
-            Position workCoord, Double feedSpeed, Double spindleSpeed,
-            OverridePercents overrides, Position workCoordinateOffset,
-            EnabledPins pins, AccessoryStates states) {
+    public ControllerStatus(String stateString, ControllerState state, Position machineCoord,
```
breiler: Codecov complains about too many arguments. The stateString-argument is temporary until all usages of that field is replaced. I don't have any good ideas on how to solve this, a builder?

winder: I've wondered the same thing: [GrblController.java#L173](https://github.com/winder/Universal-G-Code-Sender/blob/master/ugs-core/src/com/willwinder/universalgcodesender/GrblController.java#L173)

I'm a big fan of `Lombok` which has a `@Builder` annotation but have been hesitant to bring it into the project since it requires IDE plugins.

These objects are only created in a small number of places, so I don't mind ignoring the warning.

PR #999, opened 2018-04-09
Visualizer actions

winder:
New actions registered with the platform for the new camera location presets. Once again that script which generates all the light/dark icons in different sizes was invaluable.

Moved the export script so I could use it in another module.

I was very surprised to see that the `TopComponentLocalizer` worked to localize the action in the toolbar customizer and the keybindings UI. I think I'm going to rename `TopComponentLocalizer` and update the `ugs-platform-ugscore` actions to use it so that we don't need that `resources.MessagesBundle` file.

In this screenshot I manually created a toolbar with the customizer. These are probably useful enough to be enabled by default though:
visualizer_toolbar


ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/actions/CameraResetPreset.java
```diff
@@ -0,0 +1,73 @@
...
+                position = 1060)
+})
+public final class CameraResetPreset extends MoveCameraAction {
+    public static final String ICON_BASE = "icons/XYZ.png";
+    public static final String CATEGORY = LocalizingService.CATEGORY_VISUALIZER;
+    public static final String ID = "com.willwinder.ugs.nbm.visualizer.actions.CameraResetPreset";
+    public static final String NAME = Localization.getString("platform.visualizer.popup.presets.reset");
+
+    @OnStart
+    public static class Localizer extends TopComponentLocalizer {
```
winder: The `ugs-platform-ugscore` actions all set the `displayName` using a module-local MessagesBundle file. Turns out there's no need for that using this `@OnStart` helper!

breiler: Nice, I'll start using it for the actions!

Issue #993, opened 2018-04-07
Workflow Helper Plugin Finished Check Box Not Working


kenfritz:
Problem description:
The latest releases of UGS Platform (Build #997) the Workflow Helper Plugin has stopped showing the Checked box under the Finished Column after a file has completed being carved. Also, the Workflow Event Popup box no longer appears after a completed carve. This function was working as of Build #979.

Here is what the Workflow Helper looks like after the toolpath file is run now.
![workflow helper fail](https://user-images.githubusercontent.com/11965304/38449538-1c344874-39c5-11e8-8359-a19b20224704.PNG)

Expected Behavior:
Below capture shows what Build# 979 showed after the toolpath is run. Finished column showed Checked box and Popup workflow message.
![workflow working](https://user-images.githubusercontent.com/11965304/38449570-694a3d9e-39c5-11e8-9cd1-a50e99aefc04.PNG)

Steps to Reproduce the Problem
  1. Open Workflow Helper
  2. Open toolpath file
  3. Run toolpath
  4. Notice no check after completed toolpath is run
 
 Version:
 - UGS Platform 2.0 - Nightly Build 997 Not working
-  UGS Platform 2.0 - Nightly Build 979 working
 
Operating system:
  Windows 10

Platform:
X-carve
 GRBL 1.1f

breiler:
I think I've introduced the problem with this commit: a411cee2fd81401a6ac08eaabdc0f32d12513602
The method `GUIBackend. isSendingFile()` is reporting that the file is finished sending before the comm-state is changing to COMM_IDLE. This will cause the workflow helper to never detect if a file is completed:

WorkflowWindowTopComponent.UGSEvent:
```java
        if (cse.isStateChangeEvent()) {
            if (wasSending && cse.getControlState() == ControlState.COMM_IDLE)
               this.completeFile(backend.getGcodeFile());
            wasSending = backend.isSendingFile();
        }
```

I'll have a look at it!

breiler:
If this fix is accepted this would fix your issue but introduce another. =/

It will report the file as completed when all commands have been sent to the controller, not when all commands have been finished processing. 

So a dialog would appear with the file complete message but the file would still be running in the background. I will attempt to fix this in issue #989.

![screenshot 2018-04-07 18 00 28](https://user-images.githubusercontent.com/8962024/38457208-a7813fcc-3a8d-11e8-810d-cb43ffc5afca.png)

Issue #989, opened 2018-04-04
Send status panel problems in UGS Platform


breiler:
## Problem description 

There are a couple of small issues with the send progress panel that I've started working on. But the code changes grew out of control, so this issue is created to have them organised and to discuss the solutions.

### Behavior
* [ ] When pausing a job the **Duration** timer doesn't stop in the status panel. I would expect the duration timer to be paused/resumed together with the job.
![screenshot 2018-04-04 18 13 21](https://user-images.githubusercontent.com/8962024/38320150-104f6152-3834-11e8-8802-af4b82a8bf19.png)
* [ ] When the job is finished the **Sent row** count is wrong. I expect that the number of commands in the file is the same number processed.
![screenshot 2018-04-04 18 16 30](https://user-images.githubusercontent.com/8962024/38320290-76f5f538-3834-11e8-9c82-be98e31cd3d8.png)
* [ ] When the panel is made wider it can't be shrunk again.

#### Version

UGS Platform 2.0 - development branch

#### Operating system

Mac OSX

#### Platform

GRBL 1.1f

breiler:
The problem with the wrong number of sent rows exists because of two problems:
1. It counts sent rows in the Controller not rows in processed in the stream.
1. It stops the status panel update timer before all commands has finished because the finished file event is triggered when the send buffer is empty in `AbstractController.java`
```java
    public void checkStreamFinished() {
        if (this.isStreaming() && !this.comm.areActiveCommands() && this.comm.numActiveCommands() == 0) {
            String streamName = "queued commands";
            boolean isSuccess = (this.errorCount == 0);
            this.fileStreamComplete(streamName, isSuccess);

            // Make sure the GUI gets updated when the file finishes
            this.dispatchStateChange(getControlState());
        }
    }
```

What do you think of using fetching the values from streamCommands instead of trying to store command counters in the AbstractController:
```java
    public int getRowStat(RowStat stat) {
        switch (stat) {
            case TOTAL_ROWS:
                return this.numCommands;
                //return streamCommands.getNumRows(); <-- reactivate this part of the code
            case ROWS_SENT:
                return this.numCommandsSent;
                //return streamCommands.getNumRows() - streamCommands.getNumRowsRemaining(); <-- reactivate this part of the code
            case ROWS_REMAINING:
                return this.numCommands - this.numCommandsCompleted - this.numCommandsSkipped;
                //return streamCommands.getNumRowsRemaining(); <-- reactivate this part of the code
[...]

PR #1000, opened 2018-04-09
Fixed a couple of issues in send stats

breiler:
Fixes the issues with wrong statistics in the send status panel, issue #989 
Will now wait till the controller is idle before sending an event that the file stream is completed, issue #993

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
@@ -429,6 +429,7 @@ public abstract class AbstractController implements SerialCommunicatorListener, 
     private enum RowStat {
         TOTAL_ROWS,
         ROWS_SENT,
+        ROWS_COMPLETED,
         ROWS_REMAINING
     }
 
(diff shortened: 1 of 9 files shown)


ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
```diff
@@ -684,15 +690,18 @@ private synchronized void flushQueuedCommands() {
 
     // Reset send queue and idx's.
     private void flushSendQueues() {
-        this.errorCount = 0;
-        this.numCommands -= this.getRowStat(RowStat.ROWS_REMAINING);
+        errorCount = 0;
```
breiler: The error count is never used so I was tempted to remove its usage. But if it's implemented it could be pretty useful.

ugs-core/src/com/willwinder/universalgcodesender/GrblController.java
```diff
@@ -261,6 +261,7 @@ else if (GrblUtils.isGrblStatusString(response)) {
                 verbose = true;
                 
                 this.handleStatusString(response);
+                this.checkStreamFinished();
```
breiler: We now need to check if a file stream has finished when the controller state changes.

ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
```diff
@@ -743,7 +751,11 @@ public void communicatorPausedOnError() {
     }
 
     public void checkStreamFinished() {
-        if (this.isStreaming() && !this.comm.areActiveCommands() && this.comm.numActiveCommands() == 0) {
+        if (this.isStreaming() &&
+                !this.comm.areActiveCommands() &&
+                this.comm.numActiveCommands() == 0 &&
+                rowsRemaining() <= 0 &&
+                (getControlState() == COMM_IDLE || getControlState() == COMM_SENDING_PAUSED)) {
```
winder: When does `COMM_SENDING_PAUSED` need to be checked?

winder: Oh yeah, I think that was for Pre-Feed-Hold grbl where pause just stopped sending commands and let the stream proceed to an idle state.

breiler: So it's ok to merge?

winder: yes

Issue #977, opened 2018-03-28
Please add simple math features for setting work coordinates (or any other numeric input box)


Harvie:
## Feature request

Hello, it would be super useful if it was possible to manual set work coords (DRO) not only as number, but also as expression. Especialy dividing by two is very useful to find center of holes, etc... So i can set coords to `10/2` instead of just `5`. I know that i can open external calc program and compute it myself, but this would be much more user friendly. This might be useful even in other numeric inputs of ugsplatform, but especially in DRO box. In CAD/CAM software it's common practice, that any number can be entered as math formula.

How it would work:

  1. find one edge of the part
  2. zero DRO
  3. jog to oposed edge of part
  4. append `/2` to DRO and hit enter
  5. return to zero (BAM! you are in the center of the part)
  6. possibly do the same for other axes

You can see the example in this video (starting at 5:00) https://youtu.be/nkYpKWbA2P0?t=300

PR #1004, opened 2018-04-10
Set work coordinate

breiler:
Added functionality for setting work coordinates requested in #977.

It's also possible to use simple mathematical expressions such as the following:
set_work_coords

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
@@ -187,7 +187,12 @@ public abstract class AbstractController implements SerialCommunicatorListener, 
     public void resetCoordinateToZero(final Axis coord) throws Exception {
         throw new Exception(Localization.getString("controller.exception.reset"));
     }
-    
+
+    @Override
+    public void setWorkPosition(Axis axis, double position) throws Exception {
+        throw new Exception(Localization.getString("controller.exception.setworkpos"));
+    }
+
     /**
      * Disable alarm mode and put device into idle state, throw an exception 
      * if not supported.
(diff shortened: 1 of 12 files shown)


ugs-core/src/com/willwinder/universalgcodesender/model/GUIBackend.java
```diff
@@ -808,7 +812,33 @@ public void sendMessageForConsole(String msg) {
...
     }
-    
+
+    @Override
+    public void setWorkPosition(Axis axis, double position) throws Exception {
+        controller.setWorkPosition(axis, position);
+    }
+
+    @Override
+    public void setWorkPositionUsingExpression(final Axis axis, final String expression) throws Exception {
```
winder: Wow, this is really cool. I had no idea you could do something like this.

jahnj0584:
Wicked. How complex of equations can you do? I assume it only modifies the
current WCS?

winder:
If the expression starts with +, -, * or / the current value is inserted to the left, otherwise it evaluates whatever you put in directly.

So if the axis is at 10, "+ 2" is evaluated as "10 + 2". But if you type in "2 + 2" it evaluates to 4 even though you are at 10.

The scripting language is javascript so you can put whatever you want in there. I tried the `<<` bitshift operator and it worked, you can chain them together so if you are at 11, `* 1 << 1` it evaluates as `11*1<<1` which turns into 22.

jahnj0584:
But you could do something like "/2 -29" when you want it to be 29mm away
from the center?

flounderscore:
How can I set the current position to a negative value? It seems like it
would attempt do perform math instead.

winder:
@jahnj0584 it would depend on your order of operations, but I think so. Let us know, I just merged it in :)

winder:
@flounderscore you're right, you would have to do `10 * -1`

winder:
@flounderscore or zero out the axis first so that the math is right.

flounderscore:
Any chance to make the math feature optional? All my coordinates are typically <0.

ugs-core/src/com/willwinder/universalgcodesender/model/GUIBackend.java
breiler: Me neither, google was my friend! 😄

ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
```diff
@@ -187,7 +187,12 @@ public void resetCoordinatesToZero() throws Exception {
     public void resetCoordinateToZero(final Axis coord) throws Exception {
```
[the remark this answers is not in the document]
breiler: That is a good idea!

Issue #1009, opened 2018-04-15
Setup wizard


breiler:
## Feature request
I would like to see a setup wizard that would make it easier to get started with UGS configure a controller. 

I've started the implementation on a minimum viable product and would like your input. 

* Connecting to GRBL:
![screenshot 2018-04-15 21 38 19](https://user-images.githubusercontent.com/8962024/38782625-a99c00de-40f6-11e8-8c1d-003a1a46c314.png)
![screenshot 2018-04-15 21 38 37](https://user-images.githubusercontent.com/8962024/38782627-ad10c916-40f6-11e8-98d8-960ee74a6379.png)
* Motor wiring
![screenshot 2018-04-15 21 38 59](https://user-images.githubusercontent.com/8962024/38782680-850b3cac-40f7-11e8-80cd-2da46d376c7f.png)
* Calibrating step lengths
![screenshot 2018-04-15 21 39 11](https://user-images.githubusercontent.com/8962024/38782703-cbfc1384-40f7-11e8-9d4d-7903ee3dc8ee.png)
* Configuring limit switches
![screenshot 2018-05-03 14 45 29](https://user-images.githubusercontent.com/8962024/39577163-cdc70178-4ee0-11e8-810b-c585033abad2.png)
* Configuring homing
![screenshot 2018-05-03 14 46 07](https://user-images.githubusercontent.com/8962024/39577133-bfa4acb2-4ee0-11e8-8878-b99c4769cbe9.png)
* Configuring soft limits
![screenshot 2018-04-15 21 40 16](https://user-images.githubusercontent.com/8962024/38782721-ee2fcfae-40f7-11e8-924b-89c7375d3516.png)

vMeph:
great job!!!

RecceDG:
I'd like to see a UI button that is "set soft limit on [X|Y|Z] axis here"

breiler:
@RecceDG you mean a button somewhere more accessible than in the last step of a setup wizard, like in the toolbar menu?

There is a problem that, in order to use soft limits you need to have homing enabled (assuming you are using a GRBL based controller). And now we are just talking about enable/disable. There is a lot of other parameters that should be available eventually; direction, debounce, pull-off, seek rate and feed rate.

A prerequisite to enabling homing is that you need hard limits enabled.

So there is a lot of stuff that needs to be configured before being able to enabling soft limits. 
How should such a button behave if the requirements for enabling soft limits isn't met?

[...]

PR #1012, opened 2018-04-17
Feature/setup wizard

breiler:
Implementation of #1009

Added the setup wizard which can be activated under the menu "Machine/Setup wizard...". There is still a lot of work to be done, translations for instance. But I would like for it to be code reviewed and tested early.


breiler:
1. Great input. The wizard uses a step iterator which could be manipulated to add/remove/replace step with specific implementations. I'll give it some thought on how to use some sort of strategy pattern after connecting to the machine.

2. It's a good idea with a flag, maybe create a capability where we only enable it to the machines that we've tested?

3. I was thinking of using your localization service, but wanted the features to stabilize first. I think that we will find a lot of things that needs to be changed once users are able to test it.

ugs-platform/ugs-platform-plugin-setup-wizard/src/main/java/com/willwinder/ugs/nbp/setupwizard/panels/WizardPanelHardLimits.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    @Override
+    public void destroy() {
+        getBackend().removeUGSEventListener(this);
+    }
+
+    @Override
+    public boolean isEnabled() {
```
breiler: Each step is now responsible for letting the wizard know if the step is applicable for the controller.

breiler:
Is this ok to merge?

winder:
Sorry for the delay. Looking forward to people trying this out!

breiler:
No worries, I prefer that you have the time to check it through than rushing it.
Looking forward to it as well! =)

ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
```diff
@@ -240,12 +241,16 @@ static protected String getViewParserStateCommand(final double version, final Ch
...
+        ret.addCapability(CapabilitiesConstants.HOMING);
+        ret.addCapability(CapabilitiesConstants.HARD_LIMITS);
+        ret.addCapability(CapabilitiesConstants.SOFT_LIMITS);
 
         // Check if real time commands are enabled.
         if (version==0.8 && (letter != null) && (letter >= 'c')) {
             ret.addCapability(GrblCapabilitiesConstants.REAL_TIME);
         } else if (version >= 0.9) {
             ret.addCapability(GrblCapabilitiesConstants.REAL_TIME);
+            ret.addCapability(CapabilitiesConstants.SETUP_WIZARD);
```
winder: The capabilities flags look to be working nicely!

ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
```diff
@@ -240,12 +241,16 @@ static protected String getViewParserStateCommand(final double version, final Ch
     static protected Capabilities getGrblStatusCapabilities(final double version, final Character letter) {
         Capabilities ret = new Capabilities();
         ret.addCapability(CapabilitiesConstants.JOGGING);
+        ret.addCapability(CapabilitiesConstants.HOMING);
+        ret.addCapability(CapabilitiesConstants.HARD_LIMITS);
+        ret.addCapability(CapabilitiesConstants.SOFT_LIMITS);
```
winder: I don't think these are supported pre-v8 so they could be moved into one of the conditions below. Maybe change the conditions so that they are 2 if blocks instead of if/else.

ugs-core/src/com/willwinder/universalgcodesender/model/UGSEvent.java
```diff
@@ -176,4 +191,9 @@ public ControllerStatus getControllerStatus() {
     public GcodeState getGcodeState() {
       return gcodeState;
     }
+
+    @Override
+    public String toString() {
+        return ToStringBuilder.reflectionToString(this);
```
winder: I've never seen this before, nice trick.

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerListener.java
```diff
@@ -39,7 +40,14 @@ This file is part of Universal Gcode Sender (UGS).
...
      */
     void fileStreamComplete(String filename, boolean success);
-    
+
+    /**
+     * If an alarm is received from the controller
+     *
+     * @param alarm the alarm received from the controller
+     */
+    void receivedAlarm(Alarm alarm);
```
winder: Do you think this is needed in addition to `UGSEvent`? I sort of consider this class to be semi-deprecated, mainly because every time an event is added so many other classes break.

ugs-core/test/com/willwinder/universalgcodesender/firmware/GrblFirmwareSettingsTest.java
```diff
@@ -227,4 +227,137 @@ public void setValueShouldTimeoutIfNoResponseFromController() throws Exception {
...
+        assertEquals(true, target.isInvertDirectionZ());
+
+        target.rawResponseListener("$3=7");
+        assertEquals(true, target.isInvertDirectionX());
+        assertEquals(true, target.isInvertDirectionY());
+        assertEquals(true, target.isInvertDirectionZ());
+    }
+
+    @Test
+    public void setInvertDirectionXShouldSetBit() throws FirmwareSettingsException, InterruptedException {
```
winder: I'm having a hard time following some of these. At one point I used `GUIHelpers.invokeLater` to optionally run things on the same thread to make unit testing a little easier. Do you think something like that could be done here? Perhaps move that function to `ThreadHelpers`.

Another thing I've done in other Controller/Communicator classes is to have a constructor which injects the communicator so that I can mock that piece for the validations.

ugs-core/src/com/willwinder/universalgcodesender/model/UGSEvent.java
breiler: I know, what other gems could be hiding in Apache Commons? =)

These are my other favorites:
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/HashCodeBuilder.html
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/EqualsBuilder.html

ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
breiler: I went through the old documentation and homing + hard limits looks like they where developed in 0.8. Soft limits in 0.9. Could this be an option:

```java
    static protected Capabilities getGrblStatusCapabilities(final double version, final Character letter) {
        Capabilities ret = new Capabilities();
        ret.addCapability(CapabilitiesConstants.JOGGING);

        if (version >= 0.8) {
            ret.addCapability(CapabilitiesConstants.HOMING);
            ret.addCapability(CapabilitiesConstants.HARD_LIMITS);
        }

        if (version==0.8 && (letter != null) && (letter >= 'c')) {
            ret.addCapability(GrblCapabilitiesConstants.REAL_TIME);
        }

        if (version >= 0.9) {
            ret.addCapability(GrblCapabilitiesConstants.REAL_TIME);
            ret.addCapability(CapabilitiesConstants.SOFT_LIMITS);
            ret.addCapability(CapabilitiesConstants.SETUP_WIZARD);

        }

        if (version >= 1.1) {
            ret.addCapability(GrblCapabilitiesConstants.V1_FORMAT);
            ret.addCapability(GrblCapabilitiesConstants.HARDWARE_JOGGING);
            ret.addCapability(CapabilitiesConstants.OVERRIDES);
            ret.addCapability(CapabilitiesConstants.CONTINUOUS_JOGGING);
        }

        return ret;
    }
```

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerListener.java
breiler: Yeah, it got pretty messy to add this...

I couldn't find any other ways to notify the UGSEventListeners from the controller. Any spontaneous suggestions? Otherwise I can give it some thought.

ugs-core/test/com/willwinder/universalgcodesender/firmware/GrblFirmwareSettingsTest.java
breiler: Yes, that's a bit cleaner, thanks!

ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
winder: Yes, I think this looks good. The matrix might get more interesting once GRBL-ARM is done, and may even involve querying GRBL for what capabilities are even enabled.

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerListener.java
winder: That's why I say semi-deprecated. Having the Controller <> Backend communication use this is fine (I'm not sure it makes sense for `UGSEvent` to originate in the Controller. But everything outside the Backend should be accessible from UGSEvent.

PR #1014, opened 2018-04-20
Added units to pendant html page

BarrettLowe:
Hello. Please take a look here. I think something else may be wrong because the test that I wrote doesn't pass BUT for the life of me, I can't figure out why. Apparently the HTML GET status is 500 when the test is run but when I manually start everything up and use python to manually complete the test, I get a status 200. What am I missing?

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/pendantui/PendantConfigBean.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/pendantui/PendantConfigBean.java
@@ -5,6 +5,7 @@ import java.util.List;
 
 public class PendantConfigBean {
 	private List<StepSizeOption> stepSizeList = new ArrayList<>();
+	private List<UnitAmtOption> unitList = new ArrayList<>();
 	private List<ShortCutButton> shortCutButtonList = new ArrayList<>();
 	
 	{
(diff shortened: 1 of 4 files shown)


ugs-core/src/com/willwinder/universalgcodesender/pendantui/PendantUI.java
```diff
@@ -247,9 +247,21 @@ public void handle(String target,Request baseRequest,HttpServletRequest request,
                 int dirY = parseInt(baseRequest.getParameter("dirY"));
                 int dirZ = parseInt(baseRequest.getParameter("dirZ"));
                 double stepSize = parseDouble(baseRequest.getParameter("stepSize"));
+                String unitStr = baseRequest.getParameter("unitAmt");
```
winder: There's a helper for this in `UnitUtils.getUnit(String abbrev)`

ugs-core/src/com/willwinder/universalgcodesender/pendantui/PendantConfigBean.java
```diff
@@ -55,6 +59,41 @@ public String toString() {
 			return "StepSizeOption [value=" + value + ", label=" + label + ", selected=" + selected + "]";
 		}
 	}
+
+	public static class UnitAmtOption{
```
winder: I'm not totally clear on the purpose of this, would it be possible to cache just the current Unit instead of a list?

winder: Nevermind, I see that this is the pattern used by the other settings as well.

ugs-core/src/com/willwinder/universalgcodesender/pendantui/PendantConfigBean.java
```diff
@@ -13,6 +14,9 @@
 		stepSizeList.add(new StepSizeOption("5", "5", false));
 		stepSizeList.add(new StepSizeOption("10", "10", true));
 		stepSizeList.add(new StepSizeOption("50", "50", false));
+
+		unitList.add(new UnitAmtOption("MM", "MM", true));
+		unitList.add(new UnitAmtOption("IN", "IN", false));
```
winder: For the value, you should be able to use `Units.MM.abbreviateion` and `Units.INCH.abbreviation`. That way you can use the deserialize helper later on.

breiler:
Closing this as the new pendant is in place.

Issue #1013, opened 2018-04-17
Jog from Visualizer


flounderscore:
## Feature request
When initiated from visualizer, the jog should happen (IMHO) in absolute mode (G90).

## Problem description 
The gcode files created by my post processor leave the GRBL controller in G91. Therefore, jogging initiated by right-clicking the visualizer moves relative to its current (random) position.

### Expected Behavior
Switch to G90 before G0-ing.

### Actual Behavior
Leave G90/91 untouched and move relatively.

### Steps to Reproduce the Problem
  1. G91
  2. Jog via visualizer

#### Version
  UGS Platform 2.0 - Github 4/16/2018

#### Other
I'm a little torn about this "issue". It's mainly caused by a buggy post-processor, but I do think that clicking in the visualizer should cause the motion in absolute coordinates.

PR #1015, opened 2018-04-24
Always jog in absolute (G90) mode when initiated from the visualizer

flounderscore:
Attempt to solve issue #1013.


winder:
@flounderscore thanks for the contribution! The parser state is a new feature, and it's pretty cool that you can get so close to a solution with it. I think there might be race conditions with this approach, because of that the controller actually has a cache of the expected modal states and there is a helper to use that cache:
```
        backend.sendCommand(true, "G90 G0 X" + x + " Y" + y);
```

Please switch over to this approach and I'll merge in the PR.

flounderscore:
Should be good to go now (assuming I understood your comment correctly).

Issue #1016, opened 2018-04-25
MacOS Port not in list for CNC machines using CH340 chips


MarkiiD:
## Problem description 
I have a cheap Chinese Woodpecker CNC machine that uses the CH340 serial/USB chip. When I plug it in to my Mac it doesn't appear in the drop down port list. I can enter the port name manually and connect successfully.

### Expected Behavior
The serial port list should show the CNC port on the list

### Actual Behavior
The serial port isn't shown.

### Steps to Reproduce the Problem
1. Start UGS
2. Plug in the CNC machine
3. Refresh the port list
4. An error dialog pops up 'No serial ports found'

#### Version
Universal Gcode Platform (Version 2.0 [nightly] / Jul 23, 2017)

I've also downloaded the latest code and built it - I get the same results

#### Operating system
Mac OSX High Sierra

#### Platform
GRBL 0.9j

#### Other
I tried the Arduino app and the port appears, shown in the list as **/dev/cu.wchusbserial1440** so I went and downloaded the Arduino source code... they appear to be using a different version of jssc/SerialPortList.java (looks like they've copied it in and updated it).

The commit comment for this change seems to be exactly what I'm hoping can be fixed for UGS

https://github.com/arduino/Arduino/commit/f91670e0d039a237ecf73ad902bd5a487308ea64

"Fixed JSSC upload problem with Arduino Due on MacOSX
JSSC, on unix based systems like linux and MacOSX, when listing serial ports
tries to open each port to ensure its existence. While this check works well for
linux ports /dev/ttyS0..31, it leads to unexpected behaviuors on MacOSX in
particular with USB-CDC virtual serial ports.
This patch disable the check and keep it enabled only for linux ttySxx ports.

This adds also tty.* and cu.* to the list of available serial ports on MacOSX."

winder:
I've known about this issue for a while but have been reluctant to pull in the additional dependency. Unfortunately, JSSC seems to be practically abandoned to there may be no other option:
https://github.com/winder/Universal-G-Code-Sender/issues/704#issuecomment-322241643
https://github.com/winder/Universal-G-Code-Sender/issues/261#issuecomment-122441168

Issue #940, opened 2018-03-09
JSSC + Java 9 + Windows


winder:
There seems to be a driver issue with JSSC + Java 9 + Windows
Defaults here: https://github.com/scream3r/java-simple-serial-connector/issues/116

UGS could use the patched version of `jssc.jar` in the LaserCut project which may resolve the issue in some cases:
https://github.com/wholder/LaserCut/blob/master/lib/jssc.jar

breiler:
Tried replacing JSSC with [jSerialComm](http://fazecast.github.io/jSerialComm/) because the project seems more active.
It seems to work when using the classic version. However I cannot get the nbm packaging to work... 
As a side note, it lists serial ports with the prefix `cu` which could fix the issue #1016.

@winder do you have any ideas why this isn't working?
https://github.com/breiler/Universal-G-Code-Sender/commit/6e1915bb85fe03a2f7caec46b36474fd258da0ea

I get the build error: 
> [ERROR] Failed to execute goal org.codehaus.mojo:nbm-maven-plugin:4.1:manifest (default-manifest) on project ugs-platform-ugslib: Execution default-manifest of goal org.codehaus.mojo:nbm-maven-plugin:4.1:manifest failed. IllegalArgumentException -> [Help 1]
> org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:nbm-maven-plugin:4.1:manifest (default-manifest) on project ugs-platform-ugslib: Execution default-manifest of goal org.codehaus.mojo:nbm-maven-plugin:4.1:manifest failed.

breiler:
Does anyone run Windows and Java 9 that can try out the new connection driver library. 

Change from JSSC to JSerialComm in the settings:
![screen shot 2018-05-29 at 14 40 06](https://user-images.githubusercontent.com/8962024/40659348-4a279b84-634e-11e8-91f6-19bcc6f0e16e.png)

And remember to refresh the list of ports:
![screen shot 2018-05-29 at 14 40 31](https://user-images.githubusercontent.com/8962024/40659367-56f7a75a-634e-11e8-84c8-54c6ca3551c7.png)

breiler:
I was able to get a hold of a Windows machine.  Unfortunately it was running 32-bit Windows 7, and I wasn't able to reproduce the problem. However it worked with both JSSC and JSerialComm. I'm wondering if we could do an experiment setting JSerialComm as a default for new users?

[...]

PR #1029, opened 2018-05-11
Added setting for changing connection driver library

breiler:
Added an option for changing connection library to JSerialComm. I couldn't get the latest version of JSerialComm to work. The default library is still JSSC and could be removed once we've verified the new library.

This might fix: #1016 and  #940

To activate, go to settings and change the driver:
screenshot 2018-05-11 08 00 31


ugs-core/src/com/willwinder/universalgcodesender/connection/ConnectionFactory.java
```diff
@@ -18,12 +18,44 @@ This file is part of Universal Gcode Sender (UGS).
...
  * @author wwinder
  */
 public class ConnectionFactory {
-    static public Connection getConnectionFor(String address, int baud) {
-        return new JSSCConnection();
+
+    private static final Logger logger = Logger.getLogger(ConnectionFactory.class.getSimpleName());
+
+    static public Connection getConnection() {
+        Settings settings = SettingsFactory.loadSettings();
```
breiler: I couldn't find a better way to get the settings, it didn't feel right to pass it as a parameter this far down the chain. If you have any suggestions I will happily change it!

winder:
Nice! Thanks.

PR #1032, opened 2018-05-13
In the JOG module, the feed speed is not taken into account (bug)

zuzu59:
It can be corrected by simply changing the g-code command G0 to G1 so that the JOG takes into account the speed of movement (Feed).


winder:
This looks fine, the tests just need to be updated

winder:
Incorporated with #1056, thanks for the help!

PR #1051, opened 2018-06-11
Feature/import export settings

breiler:
Added two buttons in the firmware dialog for importing/exporting settings. 

The intent is that a company can export setting files for their products to their customers which can be easily imported. The file contains minimal meta data to ensure the correct files are loaded.

It can also function as a backup tool for firmware settings.

The new buttons (it felt weird with the close button to the right so I took the liberty. If you don't approve I'll change it back):
screen shot 2018-06-11 at 20 43 25

When clicking the import a confirm dialog will be shown:
screen shot 2018-06-11 at 20 45 12

The file format is as following:
```json
{
  "name": "Carbide Shapeoko 3",
  "createdBy": "Sven Svensson",
  "date": "2018-06-11",
  "firmwareName": "Grbl 1.1f",
  "settings": [
    {
      "key": "$30",
      "value": "1000",
      "units": "RPM",
      "description": "Maximum spindle speed. Sets PWM to 100% duty cycle.",
      "shortDescription": "Maximum spindle speed"
    },
    {
      "key": "$31",
      "value": "0",
      "units": "RPM",
      "description": "Minimum spindle speed. Sets PWM to 0.4% or lowest duty cycle.",
      "shortDescription": "Minimum spindle speed"
    },
    ...
```


ugs-core/src/com/willwinder/universalgcodesender/uielements/firmware/FirmwareSettingsDialog.java
```diff
@@ -99,6 +115,12 @@ private void initComponents() {
         closeButton.setText("Close");
         closeButton.addActionListener(event -> closeButtonActionPerformed());
 
+        exportButton = new JButton("Export");
```
winder: I think there are localized versions for import/export.

breiler: Yes I just used the same pattern as done before where localizations are done in `initLocalization`

```java
private void initLocalization() {
    saveButton.setText(Localization.getString("save"));
    closeButton.setText(Localization.getString("close"));
    importButton.setText(Localization.getString("import"));   
    exportButton.setText(Localization.getString("export"));
```

ugs-core/src/com/willwinder/universalgcodesender/uielements/components/SettingsFileTypeFilter.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+    
+    @Override
+    public boolean accept(File f) {
+        if (f.isDirectory()) {
+            return true;
+        }
+ 
+        String extension = getExtension(f);
+        return "settings".equals(extension);
```
winder: You could also use `StringUtils.endsWith(f.getName(), ".settings")`

ugs-core/src/com/willwinder/universalgcodesender/uielements/firmware/FirmwareSettingsDialog.java
winder: I missed that. 👍

PR #1053, opened 2018-06-12
Feature/import export settings

breiler:
Added a step in the setup wizard for importing settings. This to make it easier for CNC manufacturers to have their customers setup their machines.

Select a settings file:
screen shot 2018-06-12 at 10 45 40

Display the meta data. The title "Exported settings" comes from the file and could instead be the name for the CNC machine.
screen shot 2018-06-12 at 10 45 53

Verify to overwrite the settings
screen shot 2018-06-12 at 10 46 05

A confirmation that the settings has been imported
screen shot 2018-06-12 at 10 46 14

### DIFF
--- a/ugs-platform/ugs-platform-plugin-setup-wizard/src/main/java/com/willwinder/ugs/nbp/setupwizard/panels/WizardPanelImportSettings.java
+++ b/ugs-platform/ugs-platform-plugin-setup-wizard/src/main/java/com/willwinder/ugs/nbp/setupwizard/panels/WizardPanelImportSettings.java
@@ -1,0 +1,184 @@
+package com.willwinder.ugs.nbp.setupwizard.panels;
+
+import com.willwinder.ugs.nbp.setupwizard.AbstractWizardPanel;
+import com.willwinder.universalgcodesender.firmware.FirmwareSettingUtils;
+import com.willwinder.universalgcodesender.firmware.FirmwareSettingsException;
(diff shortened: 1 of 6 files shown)


ugs-platform/ugs-platform-plugin-setup-wizard/src/main/java/com/willwinder/ugs/nbp/setupwizard/panels/WizardPanelImportSettings.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    private void loadSettingsFile(File file) {
+        if (file == null) {
+            firmwareSettingsFile = null;
+        } else {
+            try {
+                firmwareSettingsFile = FirmwareSettingUtils.readSettings(file);
+            } catch (IOException e) {
+                JOptionPane.showMessageDialog(this.getComponent(), "Couldn't read settings file: " + e.getMessage(), "Error while reading settings file", JOptionPane.ERROR_MESSAGE);
```
winder: There's a helper in `GUIHelper.displayErrorDialog` that I've been using to wrap dialogs. It's main purpose is to wrap it with `invokeLater`, so not strictly necessary here since it will already be tied to a swing event thread.

breiler: It's nice to have it consistent, thanks!

PR #1055, opened 2018-06-13
Sets the correct version string in the about dialog

breiler:
Fixes a issue where the wrong version is displayed in the about dialog, previously seen in #1017

screen shot 2018-06-13 at 09 12 25


winder:
Would this still work if the code went in `WindowTitleUpdaterService.java` instead of the branding plugin? I don't know why I never looked for this, so many incorrect versions reported over the years, great find.

breiler:
Yes it seems to work so I changed it. It was just pure luck that I found this and it should improve some bug reports. =)

PR #1056, opened 2018-06-14
In the JOG module, the feed speed is not taken into account (bug)

tjhayasaka:
this is basically "#1032 by zuzu59" + "update tests" - "delete trailing whitespaces".


winder:
Thanks for the help!

PR #1075, opened 2018-07-06
Update Help messages to be more controller agnostic

justinclift:
Just noticed the description lines in the Help section specifically refer to GRBL.  This PR just replaces "GRBL" with "the controller" so it's a bit more controller agnostic. :smile:

Note - I made these changes blindly, without testing if the result compiles ok.  Merging blindly is probably not a good idea. :grinning:


winder:
Thanks @justinclift. That's what the continuous integration servers are for :)

PR #1076, opened 2018-07-10
Add homing cycle for g2core XYZ axes

justinclift:


breiler:
The button in the UI will still be inactivated. Add this line for telling the UI that the controller has homing capabilities in the method `TinyGController#rawResponseHandler(String response)` and it should work.

```capabilities.addCapability(CapabilitiesConstants.HOMING);```

**Comment by justinclift** [comment]
Thanks will do. :smile:

**Comment by winder** [comment]
@breiler I'll leave merging this to your judgment. It seems like maybe there is some extra work to make sure homing is properly configured before setting the homing capability.

@justinclift if @breiler wants to work on some other parts first, you can always put that command in a macro to get yourself a button in the GUI before the full integration is ready.

**Comment by justinclift** [comment]
Ahhh, no worries at all.  Yeah, an optimal approach would probably be to first check if the XYZ axes each have a (non-zero) value defined for their "homing input".

Those are the `xhi`, `yhi`, and `zhi` variables.

In text mode on my local g2core Due just now:

```
g2core[mm] ok> $xhi
[xhi] x homing input              2 [input 1-N or 0 to disable homing this axis]
g2core[mm] ok> $yhi
[yhi] y homing input              4 [input 1-N or 0 to disable homing this axis]
g2core[mm] ok> $zhi
[zhi] z homing input              6 [input 1-N or 0 to disable homing this axis]
```

And the JSON mode equivalent:

```
{"yhi":n}
{"r":{"xhi":2},"f":[1,0,11]}
{"yhi":n}
{"r":{"yhi":4},"f":[1,0,11]}
{"zhi":n}
{"r":{"zhi":6},"f":[1,0,11]}
```

When further axes are added to UGS (ABC/UVW/etc), it'd probably be optimal to expand the check to also do them, and then append that axis name to the `G28.2` homing command.

That is all way more in-depth than my personal java skill level will cope with, so I'll just hang back and cheer you guys on. :grin:

breiler:
When firmware settings support is in place we can do checks to see if homing is configured. But in the meantime this will work. Users will get an error message from the hardware if homing isn't configured.

winder:
@breiler maybe we want to rethink the Capability/Controller interaction. I'm especially curious if you could share your experience for implementing a new controller. What parts did you find to be easy and where were the pain points?

Now that there are concepts of `FirmwareSettings`, `Capabilities` and `Controller`, perhaps the `Controller` can focus on the streaming protocol / creating the `FirmwareSettings` object, and let the Settings/Capabilities sort out the rest?

A few ideas for this:

1) Remove the "actions" from the controller interface and replace them with runtime annotations. They can use something like the `Overrides` enum or the `CapabilitiesConstant` to tag specific methods like `@UGSCapability(HOMING)`.

2) Move the "action" logic into the capabilities, and have some sort of capabilities processor (whether it be a new service, a method in `BackendAPI` or `AbstractController`). The processor can then first call `grblCapabilities.hasCapability(X)`, and if it is available it can then call `grblCapability.performAction(X)`. We would probably want to convert `CapabilitiesConstants` into an enum for this to work.

3) Remove the "action" logic from `Capabilities` and let them exist as they do today in the Controller, perhaps with new `canHome` style methods.

justinclift:
> When firmware settings support is in place ...

For the firmware settings support, what kind of stuff are you thinking to do?

breiler:
@winder I'll write down my experiences as soon as I get some peace and quiet. =)

@justinclift just as there is a controller implementation where you can "fill in the blanks" there is one for firmware settings. The methods in the firmware settings can be used to enable different functions, eg if soft limits are enabled and the a bounding box can be drawn in the visualiser representing the machine boundaries. But it can be a bit tricky as these settings are different between different controllers.

justinclift:
No worries.  Sounds like that will be the place for people to configure which axes are enabled, have homing switches, travel dimensions, etc.

breiler:
@winder regarding experiences implementing support for a new controller. 

It's pretty straight forward to implement a controller when you know what happens in the lower layers. For instance the `AbstractController` uses a **template method** design pattern which usually requires the implementor to know a little about when and where the method is called from the lower layers. I think that this is fine as it is today, but could be improved with some javadoc. The most tricky part to implement was figuring out which events should be triggered and when. The actions weren't difficult to implement at all as the contract/interface for these are well defined.

> Remove the "actions" from the controller interface and replace them with runtime annotations. They can use something like the Overrides enum or the CapabilitiesConstant to tag specific methods like @UGSCapability(HOMING).

I'm not sure about this. Is this going to work with methods requiring different number of parameters. How will the implementors know what those are without a well defined contract?

> Move the "action" logic into the capabilities, and have some sort of capabilities processor (whether it be a new service, a method in BackendAPI or AbstractController). The processor can then first call grblCapabilities.hasCapability(X), and if it is available it can then call grblCapability.performAction(X). We would probably want to convert CapabilitiesConstants into an enum for this to work.

I made the Capabilities as an enum at first but thought that one of the feedback you had made sense. Some controllers may have internal capabilities such as in the ```GrblCapabilitiesConstants``` which is a pretty convenient way to "feature toggle" different parts of the controller.

I do like the idea of a service which is called from a processor. This would make the controller implementation a bit smaller and more clean.

> Remove the "action" logic from Capabilities and let them exist as they do today in the Controller, perhaps with new canHome style methods.

I'm thinking that a capability could theoretically be used to hide a homing button when a controller doesn't have that function. But a "canHome" method could be misinterpreted by the developer that the function is/isn't available at a given time, i.e. when running a gcode file.

Issue #1071, opened 2018-07-02
Improve support for TinyG / g2core


breiler:
## Feature request
Improve the support for TinyG and g2core by adding or improve the support for:
* [x] Streaming file, pausing and stopping
* [x] Status report for machine coordinates
* [x] Status report for work coordinates
* [x] Handling gcode states
* [x] Jogging
* [x] Homing
* [x] Firmware settings

justinclift:
Ahhh, just came here to report "I can't seem to enable homing".  Turns out it's a known thing. :wink:

For me in testing now, using firmware version 101.03 (our latest edge build):

* Jogging is working
* Pause/resume is working
* Stopping works too
* Streaming a file seems to be working (did it with the motors off, and it's still running atm).

Homing works if done via the "Command" line (eg `G28.2 Z0 X0 Y0`), and the XYZ co-ordinates are correctly set when the home operation completes.

Noticed two bits of weirdness.  One is probably just me :wink:, while the other one might not be an issue or may need thinking about:

* "XY Step size" seems to be applied to Z movements too.  That wasn't obvious until trying it out and hitting my Z limit switch.  Should the title of that be "XYZ Step size" or is it expected to change the Z step size somewhere?
* Default feed rate (for me) was "10".  That's **super** slow.  With my Shapeoko 3 my normal feed rate is around 10-15k.  I don't remember if the feedrate is much lower with the stock Shapeoko 3 board though, as I haven't used that in ages.  It might well be.

Anyway, so far things look good.  Nothing terrible is jumping out at me anyway. :smile:

justinclift:
Hmmm, noticing one thing that I'm not sure about.  In the "Controller State (DRO)" Window, the line at the bottom looks like it's supposed to be the currently executing G-Code.  That's not changing for me atm, even though a file is being streamed.  The feed rate value and XYZ coordinates are changing dynamically though.

Should I take a screenshot and add it here?

justinclift:
#1075 updates the Help info to be more controller agnostic too.  Probably useful. :smile:

justinclift:
The file I was streaming (with motors off) finished successfully from the UGS perspective just now, so looks like that's good too. :smile:

justinclift:
> In the "Controller State (DRO)" Window, the line at the bottom looks like it's supposed to be the currently executing G-Code.

Ahhh, ignore that.  Just realised that's the "gcode state" mentioned in the welcome page tab.  No worries. :smile:

[...]

PR #1078, opened 2018-07-11
Added support for gcode states and work coordinates

breiler:
Added support for handling gcode states from the responses of the controller. It's also now possible to set and handle work coordinates. 

Further improvements for #1071

A small demo can be found here:
https://www.youtube.com/watch?v=_kmXCvtE1Pc

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/TinyGController.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/TinyGController.java
@@ -144,11 +144,6 @@ public class TinyGController extends AbstractController {
     }
 
     @Override
-    public void restoreParserModalState() {
-        // no op
-    }
-
-    @Override
     protected void rawResponseHandler(String response) {
         JsonObject jo;
 
@@ -215,8 +210,10 @@ public class TinyGController extends AbstractController {
         ControllerState previousState = controllerStatus.getState();
         UGSEvent.ControlState previousControlState = getControlState(previousState);
 
+        TinyGUtils.updateGcodeState(this, jo);
+
         // Notify our listeners about the new status
-        controllerStatus = TinyGUtils.updateControllerStatus(controllerStatus, jo);
+        controllerStatus = TinyGUtils.updateControllerStatus(controllerStatus, getCurrentGcodeState(), jo);
         dispatchStatusString(controllerStatus);
 
         // Notify state change to our listeners
@@ -244,6 +241,9 @@ public class TinyGController extends AbstractController {
         // 0=off, 1=filtered, 2=verbose
         comm.queueStringForComm("{sv:1}");
 
+        // Configure status reports
+        comm.queueStringForComm("{sr:{posx:t, posy:t, posz:t, mpox:t, mpoy:t, mpoz:t, plan:t, vel:t, unit:t, stat:t, dist:t, admo:t, frmo:t, coor:t}}");
+
         // Request initial status report
         comm.queueStringForComm("{sr:n}");
 
@@ -260,7 +260,8 @@ public class TinyGController extends AbstractController {
 
     @Override
     public void resetCoordinatesToZero() throws Exception {
-        throw new UnsupportedOperationException(NOT_SUPPORTED_YET);
+        String command = TinyGUtils.generateResetCoordinatesToZeroCommand(controllerStatus, getCurrentGcodeState());
+        sendCommandImmediately(new GcodeCommand(command));
     }
 
     @Override
@@ -291,11 +292,7 @@ public class TinyGController extends AbstractController {
 
     @Override
     public void updateParserModalState(GcodeCommand command) {
-        if (command.getCommandString().startsWith("{\"gc")) {
-            String gcode = StringUtils.substringBetween(command.getCommandString(), "{\"gc\":\"", "\"");
-            GcodeCommand gcodeCommand = new GcodeCommand(gcode);
-            super.updateParserModalState(gcodeCommand);
-        }
+        // This is handled by the rawResponseHandler
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


ugs-core/src/com/willwinder/universalgcodesender/TinyGUtils.java
```diff
@@ -206,4 +230,150 @@ private static String getStateAsString(int state) {
...
+     */
+    public static String generateSetWorkPositionCommand(ControllerStatus controllerStatus, GcodeState gcodeState, Axis axis, double position) {
+        int offsetCode = convertOffsetGcodeToCode(gcodeState.offset);
+        Position machineCoord = controllerStatus.getMachineCoord();
+        double coordinate = -(position - machineCoord.get(axis));
+        return "G10 L2 P" + offsetCode + " " +
+                axis.name() + Utils.formatter.format(coordinate);
+    }
+
+    private static int convertOffsetGcodeToCode(Code offsetGcode) {
```
winder: This could go right into GcodeUtils.java, maybe even some of the other WCS helpers you have here as well. There is also WorkCoordinateSystem.java, so something like this might technically work:
```
 WorkCoordinateSystem.valueOf(offsetGcode.toString()).getPValue();
```

winder: I'd probably lean towards keeping your static method and maybe getting rid of `WorkCoordinateSystem`

ugs-core/src/com/willwinder/universalgcodesender/TinyGController.java
```diff
@@ -292,11 +293,7 @@ public void viewParserState() throws Exception {
 
     @Override
     public void updateParserModalState(GcodeCommand command) {
-        if (command.getCommandString().startsWith("{\"gc")) {
-            String gcode = StringUtils.substringBetween(command.getCommandString(), "{\"gc\":\"", "\"");
-            GcodeCommand gcodeCommand = new GcodeCommand(gcode);
-            super.updateParserModalState(gcodeCommand);
-        }
+        // This is handled by the rawResponseHandler
```
winder: This calls `parser.addCommand` which updates the on-the-fly GcodeParser, that may need to be called somewhere. I think my plan was to deprecate distanceModeCode / unitsCode in favor of the real gcode parser.

breiler:
It seems that I've been missing a key feature by not adding the parser. =)

I figured that because the controller reports all state changes in the status report I don't need to parse the sent gcode. The way I'm updating the state felt a bit dodgy from the start so this clearly needs to be updated. 

I'll see what I can do!

breiler:
Are you referring to the machine coordinates?

Yes according to the documentation they are always in MM: https://github.com/synthetos/g2/wiki/Status-Reports

breiler:
I noticed some problems with the distance mode parsing, need to test it further later.

breiler:
There was a bit back and forth with this, but I'm starting to feel satisfied. 
Does this work for you @winder?

ugs-core/src/com/willwinder/universalgcodesender/TinyGController.java
```diff
@@ -254,14 +257,23 @@ private void sendInitCommands() {
...
+        // Prevent internal TinyG commands to update the parser modal state
+        if(!command.getCommandString().startsWith("{")) {
+            super.updateParserModalState(command);
+        }
+    }
+
     @Override
     public void performHomingCycle() throws Exception {
-        sendCommandImmediately(new GcodeCommand("G28.2 Z0 X0 Y0"));
+        sendCommandImmediately(new GcodeCommand("G28.2 X0 Y0 Z0"));
```
justinclift: As a data point, this won't make any functional difference.  The Z axis will be homed first anyway, as g2core regards that as being safer than homing it after the others.

breiler: Sorry, this is from a bad merge. It's better having the Z probe first as it will reflect the correct order in g2core, will fix this soon!

justinclift: No worries at all. :smile:

winder:
@breiler looks good to me!

justinclift:
Just tried the [newest build](https://winder.ci.cloudbees.com/job/UGS/1064/) which according to the Jenkins job info includes this... and the state change bits don't seem to be working right.

If I go to the State tab and change (say) the Units mode from G21 to G20, the "Controller state" (DRO) goes to "RUN" and keeps there.  It doesn't seem to recognise that the state change has completed.

I need to press the Cancel button in the toolbar to get the DRO back to IDLE.

Is this expected? :smile:

breiler:
I had the same issue when running g2core 101.03 which doesn't report the correct status. I posted a bug report for this in synthetos/g2#367

I could create a timer that polls the status, but it felt wrong when g2core has this status report function.

justinclift:
Yep, 100.26 is working properly.  Thanks @breiler. :smile:

justinclift:
Found what seems like a bug in this, when testing it out today on a Win 7 x64 PC.  Although the state change bit works, the "Home Machine" button is greyed out.

Maybe another side effect of the bad merge?

justinclift:
Oh that's weird.  Now the "Home Machine" button isn't being greyed out.  Same Win7 PC, but it's been restarted.  For the past few hours (actually using UGS to cut stuff), I've been having to manually type G28.2 (etc) to home things.

I'll keep an eye on this, and let you know if the homing button starts playing up again.

justinclift:
Cutting stuff today, and no issues at all.  Guess it really was some kind of weirdness from the failing PC. :smile:

Just as a random data point, the stuff cut today was some PC expansion port brackets for network cards that came without them.  This was today's (in 0.75mm steel):

dscn0130_v1

And when mounted (the one on the right).  The one on the left was yesterday's work, done in 0.75mm aluminium:

dscn0134

UGS seems to be working pretty well for this already. :smile:

breiler:
This is awesome, thanks for showing this! 😃

PR #1085, opened 2018-07-17
Add TinyG/g2core to README.md

justinclift:
Just noticed the README is still GRBL specific.  This adds TinyG/g2core to the intro paragraph as well. :smile:


justinclift:
Oops.  Just realised the first instances of TinyG & g2core should probably have been links to the projects on GitHub, like the first instance of GRBL is.  Extra PR coming up. :wink:

PR #1091, opened 2018-07-23
Remove old jog control component

### LINKED ISSUE #1089: Make the jog plugin the default jog controller
The new jog plugin supports continuous jogging, dark theme and is better adjusted to the given window size. 

Make it the default jog module and remove the old.

We've also got some support issues where users want separate step sizes for XY and Z. Should we reenable this as the default again?

breiler:
Activated the new jog module by default. Also activated separate step settings for XY  and Z by default.

If we remove the old module it would have the side effect that old users will have the old jog controller completely removed without having the new activated. So I added a check for when the old module is loaded, it will try to load the new then close it self. 

We need to keep the old module a couple of months to make sure people have transitioned to the new before we can remove it.

Fixes #1089

### DIFF
--- a/ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/JogTopComponent.java
+++ b/ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/JogTopComponent.java
@@ -53,8 +53,8 @@ import javax.swing.JPopupMenu;
         preferredID = "JogTopComponent"
 )
 @TopComponent.Registration(
-        mode = "top_left",
-        openAtStartup = false)
+        mode = "middle_left",
+        openAtStartup = true)
 @ActionID(
         category = JogTopComponent.CATEGORY,
         id = JogTopComponent.ACTION_ID)
--- a/ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/control/JogControlTopComponent.java
+++ b/ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/control/JogControlTopComponent.java
@@ -38,12 +38,7 @@ import org.openide.windows.WindowManager;
 @TopComponent.Description(
         preferredID = "JogControlTopComponent",
         //iconBase="SET/PATH/TO/ICON/HERE", 
-        persistenceType = TopComponent.PERSISTENCE_ALWAYS
-)
-@TopComponent.Registration(mode = "middle_left", openAtStartup = false)
-@TopComponent.OpenActionRegistration(
-        displayName = "<Not localized:JogControlTopComponent>",
-        preferredID = "JogControlTopComponent"
+        persistenceType = TopComponent.PERSISTENCE_NEVER
 )
 public final class JogControlTopComponent extends TopComponent {
     public JogControlTopComponent() {
--- a/ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/control/JogControlTopComponent.java
+++ b/ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/control/JogControlTopComponent.java
@@ -28,18 +28,19 @@ import java.awt.BorderLayout;
 import org.openide.awt.ActionID;
 import org.openide.awt.ActionReference;
 import org.openide.windows.TopComponent;
+import org.openide.windows.WindowManager;
 
 /**
- * Top component which displays something.
+ * Old jog controller
+ *
+ * @deprecated This jog controller is deprecated and should be replaced by the new jog module.
  */
 @TopComponent.Description(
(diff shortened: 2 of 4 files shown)


ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/JogTopComponent.java
```diff
@@ -53,8 +53,8 @@ This file is part of Universal Gcode Sender (UGS).
         preferredID = "JogTopComponent"
```
winder: This is missing the `persistenceType = TopComponent.PERSISTENCE_ALWAYS ` property, I think setting that helps the UI load faster.

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/control/JogControlTopComponent.java
```diff
@@ -28,18 +28,19 @@ This file is part of Universal Gcode Sender (UGS).
...
 /**
- * Top component which displays something.
+ * Old jog controller
+ *
+ * @deprecated This jog controller is deprecated and should be replaced by the new jog module.
  */
 @TopComponent.Description(
         preferredID = "JogControlTopComponent",
         //iconBase="SET/PATH/TO/ICON/HERE", 
         persistenceType = TopComponent.PERSISTENCE_ALWAYS
```
winder: You can update this to `PERSISTENCE_NEVER`

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/control/JogControlTopComponent.java
```diff
@@ -28,18 +28,19 @@ This file is part of Universal Gcode Sender (UGS).
...
 )
-@TopComponent.Registration(mode = "middle_left", openAtStartup = true)
-@ActionID(category = LocalizingService.JogControlCategory, id = LocalizingService.JogControlActionId)
-@ActionReference(path = LocalizingService.JogControlWindowPath)
+@TopComponent.Registration(mode = "middle_left", openAtStartup = false)
 @TopComponent.OpenActionRegistration(
```
winder: This annotation creates the actions in the Window menu, so they should be removed too. Removing old actions has been a big challenge for me, I think you're supposed to do it in `layer.xml`, there are a few examples here:

./ugs-platform/ugs-platform-ugscore/src/main/resources/com/willwinder/ugs/nbp/core/layer.xml

Details here: http://wiki.netbeans.org/NbmIdioms#XML_Layers

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/control/JogControlTopComponent.java
```diff
@@ -56,6 +57,13 @@ public JogControlTopComponent() {
     public void componentOpened() {
         setName(LocalizingService.JogControlTitle);
         setToolTipText(LocalizingService.JogControlTooltip);
+
+        // Try to load the new jog module
+        TopComponent jogTopComponent = WindowManager.getDefault().findTopComponent("JogTopComponent");
+        if(jogTopComponent != null && !jogTopComponent.isOpened()) {
+            jogTopComponent.open();
+            close();
+        }
```
winder: Hopefully disabling the module in the layer.xml makes it so you don't need to do this. Did the new jog component fail to load after changing `openAtStartup` to true?

ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/JogTopComponent.java
breiler: This is the default value so I didn't bother setting it. Would you like to have it set anyway?

TopComponent:
```java
public static @interface Description {
        /** Default value for {@link TopComponent#getPersistenceType()}. 
         * 
         * @return one of {@link TopComponent#PERSISTENCE_ALWAYS}, 
         * {@link TopComponent#PERSISTENCE_NEVER}, 
         * {@link TopComponent#PERSISTENCE_ONLY_OPENED}
         * 
         */
        public int persistenceType() default PERSISTENCE_ALWAYS;
    }
```

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/control/JogControlTopComponent.java
breiler: Yes the new module wouldn't start after changing the `openAtStartup`. It required to be activated manually.

breiler:
Is it still ok to merge this?

winder:
Yeah, whenever you think it's ready. The new jog component is better in every way as far as I can tell :)

Issue #1099, opened 2018-07-30
Macro panel in UGS Classic is invisible


breiler:
## Problem description 
The macros panel in UGS classic is invisible. Also reported here: https://groups.google.com/d/msg/universal-gcode-sender/JK9LSp_wuZk/E9Q57O0ICgAJ

#### Version
UGS Classic Nightly build

#### Operating system
MacOSX

#### Platform
GRBL 1.1f

PR #1100, opened 2018-07-30
Fixes invisible macros panel

breiler:
Fixes issue #1099

Somehow the panel layout makes it invisible. When removing it seems to work.


breiler:
Realised that this could cause problems as the code is auto generated from Netbeans. Any other ideas on how to fix it?

```java
    /** This method is called from within the constructor to
     * initialize the form.
     * WARNING: Do NOT modify this code. The content of this method is
     * always regenerated by the Form Editor.
     */
    @SuppressWarnings("unchecked")
    // <editor-fold defaultstate="collapsed" desc="Generated Code">//GEN-BEGIN:initComponents
    private void initComponents() {
```

winder:
@breiler I'll get this. I threw the original UI together with the GUI builder in NetBeans. All things considered, it works pretty well for getting the basics setup so you can test the backend stuff. I still use it sometimes:

screen shot 2018-07-31 at 8 41 53 am
screen shot 2018-07-31 at 8 42 08 am

PR #1104, opened 2018-08-05
Check preferred units before setting soft limit value.

winder:
I think this fixes the MachineBoundaries problem.  The visualizer uses MM internally, so everything has to be converted.


breiler:
I'm sorry, when the problem occured it was probably because some bad state. I think it occurs when switching back and forth between $13 settings. I haven't had time to investigate.

The unit for the soft limits are always in MM and shouldn't be transformed by the visualiser. Sorry for taking up your time!

winder:
@breiler not a problem!

PR #1106, opened 2018-08-06
Added translation support for the setup wizard

### LINKED ISSUE #1058: Add translation support for the Setup Wizard
## Feature request
The Setup Wizard is in english only, add support for localization.

breiler:
Fixes #1058


winder:
Looks good to me, my only question is with the `<p>` tags. Maybe those could be added in the code so that localizers didn't have to worry about them?

PR #1117, opened 2018-09-02
Spindle dwell support for m4

jamawa:
UGS currently can insert an optional pause after a M3 command to start the spindle in clockwise rotation. The M4 command, to start the spindle counterclockwise, does not have this feature.
This pull request updates the gcode processor class that handles this optional pause to also insert the pause after a M4 command.

The class was called M3Dweller. As this no longer fully covers the function, it has been renamed to SpindleOnDweller. This results in multiple files needing this renaming. 
Tests have been added to the existing M3 test-set.

All existing entries in the language files have also been adapted to reflect the change.


winder:
Thanks for the contribution!

jamawa:
You’re welcome! And thank you for your good work on this project, much appreciated!

PR #1166, opened 2018-12-19
Autoleveler fixes

JulioMCDias:
Hello, all right? 

I made some autoleveler corrections that appeared for Min, if you can look at this, maybe I can help you.

Fixes:

-Clean autolever before applying surface leveling.

-Disable function to apply to Gcode when already applied.

-Correction of Probe collection:
	-Collect all measurement points.
	-Returns values from the job and not from the machine.

-Add checkbox to Autoleveler visible.

-inch scale.


winder:
Thanks for looking into this! With your changes are you able to use the feature in a real job?

JulioMCDias:
Yes, I tested it on a printed circuit board and it worked very well.
The firmware version tested is grbl 0.9 and grbl 1.1.

FelisucoVFR:
Hi colleges!
Many thanks for all this work.
How can I update my UGS to get these changes?

PR #1182, opened 2019-01-26
Fixed issues with ProbeModule for inch units

jonadams1010:
Hi Will,
I encountered some issues when trying to probe in inch units, so I thought I see if I could fix it. The tests pass, and I did some manual testing on my machine, and everything appears to be working correctly (in inch and mm units).

I'm new to the codebase (and actually to github as well), though I've been doing java development for a long time. Let me know what you think.

Thanks,
Jon


winder:
Looks good, thanks for the contribution.

PR #1188, opened 2019-02-07
Fixes #1185 - make sure the spinner component issues change events

### LINKED ISSUE #1185: Jog controller step size
## Problem description 
When setting a step size, if you don't select another text box, the old value will be executed instead of actually input value

### Expected Behavior
Refresh to actual value to avoid z crashes

### Actual Behavior
Could result in table crash. Did already the first time, when I typed 1 in z travel after fast approchaing with 5. Forgot to change selection box and jogged Z-. Tool ended 4 millimeter in the table

1. Open jog controller
2. Move Z down without changing jog value. Keep the step size Z in mind though
3. Change the step size value to 1 by clicking in the box, not using the arrows, using your keyboard. Do not click in another box after typing your new step size
4. Click on Z- to jog down (make sure you have space enough!!)
5. The machine will travel by the last value and not by 1!

#### Version  
  UGS Platform 2.0 - Nightly Build Dec 18, 2018

#### Operating system
  Windows 10

#### Platform
  GRBL 1.1f

breiler:
Make sure the spinner component issues change events every time the text field is modified.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/uielements/jog/StepSizeSpinner.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/uielements/jog/StepSizeSpinner.java
@@ -19,6 +19,7 @@
 package com.willwinder.universalgcodesender.uielements.jog;
 
 import javax.swing.*;
+import javax.swing.text.DefaultFormatter;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.text.ParseException;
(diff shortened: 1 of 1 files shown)


vMeph:
@breiler  where are going all the updates of UGS platform, the CloudBees dont works anymore, so im wondering where is the new URL

breiler:
Sorry, I don't think there is any solution up and running yet. I know that @winder has experimented with Artifactory which could be a good alternative. We will have to wait and see.

vMeph:
@breiler  thanks alot for info...
hopefully @winder will come up with a solution

Issue #1192, opened 2019-02-12
New pendant


breiler:
I'm working on a new web pendant implementation and would like some feedback on my work this far.
Is there any missing must-have features?

### Waiting for connection
When loosing connection with the backend server this dialog will pop up:
![screenshot 2019-02-12 19 59 14](https://user-images.githubusercontent.com/8962024/52661368-2cd35c00-2f02-11e9-8146-45dede018107.png)

###  Connection
If we aren't connected to a controller this dialog will allow us to connect:
![screenshot 2019-02-12 19 59 35](https://user-images.githubusercontent.com/8962024/52661410-44aae000-2f02-11e9-8507-e3aff520b794.png)

### Main window
Main window for controlling the machine:
![screenshot 2019-02-12 20 03 07](https://user-images.githubusercontent.com/8962024/52661455-63a97200-2f02-11e9-9588-83d015bc8e43.png)

![screenshot 2019-02-12 20 04 03](https://user-images.githubusercontent.com/8962024/52661577-a9fed100-2f02-11e9-8ac6-4590c0c56206.png)

### Open and sending
Opening a new file and start sending:
![screenshot 2019-02-12 20 03 22](https://user-images.githubusercontent.com/8962024/52661473-702dca80-2f02-11e9-8c4c-b0578b3eb58e.png)

![screenshot 2019-02-12 20 04 17](https://user-images.githubusercontent.com/8962024/52661599-b4b96600-2f02-11e9-80a3-68290904e7fd.png)

### Jog controller
Jog controller, how should the step sizes be controlled?:
![screenshot 2019-02-12 20 03 33](https://user-images.githubusercontent.com/8962024/52661523-8b98d580-2f02-11e9-8bb4-ea3e7661061c.png)

![screenshot 2019-02-12 20 04 25](https://user-images.githubusercontent.com/8962024/52661620-bbe07400-2f02-11e9-8874-e56e04c2781f.png)

### Macros
![screenshot 2019-02-12 20 03 42](https://user-images.githubusercontent.com/8962024/52661557-a0756900-2f02-11e9-840f-cdb1f11e7330.png)

![screenshot 2019-02-12 20 04 31](https://user-images.githubusercontent.com/8962024/52661638-c13dbe80-2f02-11e9-932a-eb33c0c0285c.png)

winder:
Wow this is really nice!

khturnings:
[...]

PR #1203, opened 2019-03-14
New pendant

breiler:
Sorry for this massive pull request!

The pendant is now it's own module, I think that this gets cleaner with all the web stuff happening in it. 
Needed to move the classic gui to it's own module to avoid cyclic dependencies. 

The new pendant uses Angular 7. There is a readme in the webapp root on how to get started developing. Basically you can run UGS Classic providing the API and then run a separate process for the web application during development.

 I think I've covered all the features in the old pendant and some more. The old pendant is still available under http://localhost:8080/old

There is still some error handling and localization left to do. But to prevent an even bigger pull request I think it's best to make those changes in stages.

See #1192 for screen shots.


ugs-core/src/com/willwinder/universalgcodesender/MacroHelper.java
```diff
@@ -95,22 +95,22 @@ public static void executeCustomGcode(final String str, BackendAPI backend) thro
      * @return 
      */
     protected static String substituteValues(String str, BackendAPI backend) {
-        SystemStateBean bean = new SystemStateBean();
```
breiler: We don't rely on the system state bean anymore. That logic is now moved to the pendant implementation.

ugs-core/src/com/willwinder/universalgcodesender/services/JogService.java
```diff
@@ -30,19 +31,11 @@ This file is part of Universal Gcode Sender (UGS).
  */
 public class JogService {
     private static final Logger logger = Logger.getLogger(JogService.class.getSimpleName());
```
breiler: Got wierd errors because of these state fields never got updated. Now uses the settings instead.

ugs-pendant/src/main/webapp/proxy.conf.json
```diff
@@ -0,0 +1,8 @@
+{
```
breiler: When developing on the web client, this setting will allow to connect to a running API

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/PendantUI.java
```diff
@@ -55,65 +58,82 @@
...
         configContext.setInitParameter("cacheControl", "max-age=0, public");
 
-        HandlerList handlers = new HandlerList();
-        handlers.setHandlers(new Handler[] {configContext, sendGcodeContext, adjustManualLocationContext, getSystemStateContext, resourceHandler, new DefaultHandler()});
+        // Create a servlet servletContextHandler
+        ServletContextHandler servletContextHandler = new ServletContextHandler(ServletContextHandler.NO_SESSIONS);
+        servletContextHandler.setContextPath("/api");
+        ServletHolder servletHolder = servletContextHandler.addServlet(ServletContainer.class, "/*");
+        servletHolder.setInitOrder(1);
+        servletHolder.setInitParameter("javax.ws.rs.Application", AppConfig.class.getCanonicalName());
```
winder: I see, so this is why it's difficult to inject the backend object.

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/PendantUI.java
```diff
@@ -219,11 +239,49 @@ public void handle(String target,Request baseRequest,HttpServletRequest request,
         @Override
         public void handle(String target,Request baseRequest,HttpServletRequest request,HttpServletResponse response) throws IOException, ServletException {
             baseRequest.setHandled(true);
-            mainWindow.updateSystemState(systemState);
+            updateSystemState(systemState);
             response.getWriter().print(getSystemStateJson());
         }
     }
-    
+
```
winder: I think this can be marked as `@deprecated` with a comment that it is only used by the old pendant

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/controllers/FilesController.java
```diff
@@ -0,0 +1,83 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response pause() {
+        try {
+            if (!backendAPI.isPaused()) {
+                backendAPI.pauseResume();
+            }
+
+            return Response.ok().build();
+        } catch (Exception e) {
+            return Response.serverError().build();
```
winder: You can register an exception mapper with Jersey to do this automatically, that way you can just let the exceptions bubble up without wrapping everything with a try/catch:
```
class ExceptionMapper : ExceptionMapper<Exception> {
    override fun toResponse(e: Exception): Response {
        return Response.status(Response.Status.INTERNAL_SERVER_ERROR)
                .type(MediaType.APPLICATION_JSON_TYPE)
                .entity(ErrorMessage(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(), e.message))
                .build()
    }
}
```

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/controllers/MachineController.java
```diff
@@ -0,0 +1,238 @@
...
+
+    @Inject
+    private JogService jogService;
+
+    @GET
+    @Path("connect")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response connect() throws Exception {
+        if (backendAPI.isConnected()) {
+            return Response.status(Response.Status.NOT_ACCEPTABLE).build();
```
winder: When you mentioned error handling in the PR comment, does that include things like setting the status messages in calls like this?

Maybe `backendAPI.connect` should throw a usable exception message so that can be used instead of checking them here as well?

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/controllers/SettingsController.java
```diff
@@ -0,0 +1,53 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getSettings() {
+        try {
+            com.willwinder.universalgcodesender.utils.Settings settings = backendAPI.getSettings();
+            Settings response = new Settings();
+            response.setJogFeedRate(settings.getJogFeedRate());
+            response.setJogStepSizeXY(settings.getManualModeStepSize());
+            response.setJogStepSizeZ(settings.getzJogStepSize());
+            response.setPreferredUnits(settings.getPreferredUnits());
+            return Response.ok(response).build();
```
winder: Can you just return "Settings" here? I'm not sure exactly what magic Dropwizard adds, but it has always let me return POJOs directly rather than wrap them in a Response.

winder: I also find that makes the code easier to read when the signature tells you the return type (not a big deal for such a small function)

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/model/Settings.java
```diff
@@ -0,0 +1,45 @@
+package com.willwinder.universalgcodesender.pendantui.v1.model;
```
winder: I'm a big fan of using Kotlin for POJOs, the equivalent Kotlin:
```kotlin
data class Settings(
  val jogFeedRate: double,
  val jogStepSizeXY: double,
  val preferredUnits: UnitUtils.Units,
  val jogStepSizeZ: double
)
```

However, Kotlin -> Java interop is not as clean as Java -> Kotlin, so leaving it the way you have it is probably for the best.

ugs-pendant/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <version>3.4</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.github.eirslett</groupId>
+                <artifactId>frontend-maven-plugin</artifactId>
```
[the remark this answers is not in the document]
breiler: I haven't found any issues regarding this and the plugin seems to be actively developed so I expect that it will work.

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/AppConfig.java
```diff
@@ -0,0 +1,39 @@
...
+@ApplicationPath("/api")
+public class AppConfig extends ResourceConfig {
+    public AppConfig() {
+        register(StatusController.class);
+        register(MachineController.class);
+        register(TextController.class);
+        register(MacrosController.class);
+        register(SettingsController.class);
+        register(FilesController.class);
+        register(MultiPartFeature.class);
```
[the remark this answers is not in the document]
breiler: Ah, I've been using Spring Boot lately... =)
I'll change it!

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/PendantUI.java
breiler: It was some time ago that I tried this out. But as I remember I couldn't use the constructor fetch the constructor. This was the cleanest way to get it working.

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/PendantUI.java
breiler: Done. I think that these Handlers should be deprecated and removed once we get the new Pendant verified by the users.

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/controllers/MachineController.java
breiler: I've cleaned this up a bit using exceptions suitable for this. Actually, the error handling that I mentioned is in the GUI. But I think this needs to mature a bit before adding that.

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/resources/FilesResource.java
```diff
@@ -0,0 +1,65 @@
...
+            backendAPI.pauseResume();
+        }
+    }
+
+    @GET
+    @Path("cancel")
+    @Produces(MediaType.APPLICATION_JSON)
+    public void cancel() throws Exception {
+        throw new RuntimeException("Hepp");
+        //backendAPI.cancel();
```
winder: Looks like you forgot to change this one back.

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/resources/MachineResource.java
```diff
@@ -0,0 +1,199 @@
...
+        Settings settings = SettingsFactory.loadSettings();
+        backendAPI.connect(settings.getFirmwareVersion(), settings.getPort(), Integer.valueOf(settings.getPortRate()));
+    }
+
+    @GET
+    @Path("disconnect")
+    @Produces(MediaType.APPLICATION_JSON)
+    public void disconnect() throws Exception {
+        if (!backendAPI.isConnected()) {
+            throw new NotAcceptableException("Not connected");
```
winder: This can probably be a no-op rather than an error.

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/resources/FilesResource.java
breiler: Wow, thanks hawk eyes!

ugs-pendant/src/main/java/com/willwinder/universalgcodesender/pendantui/v1/resources/MachineResource.java
breiler: I'll leave it as it is for now. This is getting fixed in the candy mechanics branch.

PR #1206, opened 2019-03-18
Add deployment of nightly build of ugsplatform to gh-pages

mike-pittelko:
This PR will add the necessary bits and bobs to .travis.yml to get the build artifacts deployed to winder/Universal-G-Code-Server branch gh-pages.  As it stands, no builds have posted since mid December 2018.

It won't work on it's own, you also need to make a change to the travis-ci settings for the respository.  In the "settings" page, add an environment variable with a github API key in it.  You can do this by generating one on github, and pasting it in to the box on travis.  It will not be displayed in any logs, and will not appear in the .yml file.

Here are the instructions on how to set up the environment variable: https://docs.travis-ci.com/user/environment-variables#defining-variables-in-repository-settings
Their example is "h" - you would use GITHUBAPIKEY
To generate the api key on github, go here: https://github.com/settings/tokens

if you want more than just ugsplatform, adjust the local-dir in the .travis.yml file.

Now, every time a build happens, the artifacts will end up in the gh-pages branch of the main repo.  This might not be the ideal way to make this all happen, but it works. Getting this hooked up to the web pages is some other exercise, sorry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: java
 
-
 cache:
   directories:
   - $HOME/.m2
@@ -16,13 +15,12 @@ after_success:
 before_script:
   - pip install --user codecov
 
+# deploy via pages on github
+
 deploy:
-  provider: releases
-  api_key: GITHUBAPIKEY
-  file: "/home/travis/build/mike-pittelko/Universal-G-Code-Sender-Build/ugs-platform/application/target/ugs-platform-app-2.0-SNAPSHOT.zip"
-  skip_cleanup: true
-  overwrite: true
+  provider: pages
+  skip-cleanup: true
+  github-token: $GITHUBAPIKEY  # Set in the settings page of your repository, as a secure variable
+  keep-history: true
   on:
-    tags: false
-    
-    
+    branch: master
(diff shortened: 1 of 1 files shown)


winder:
Do you have an example of what this would look like once deployed?

mike-pittelko:
Go here: https://github.com/mike-pittelko/Universal-G-Code-Sender-Build/tree/gh-pages
looks exactly like this.  Click the download and can bring it down as a zip file.  Is on the gh-pages branch of the base repo.  You can actually set up the base wherever you want, but its pretty large if you take a snapshot of the whole repo.

breiler:
Sorry, I not a fan of this solution. I don't think that this will scale well when it's time to provide stable releases. 

As for the "download zip", it takes some time for github to generate the zip-file and we have many downloads daily which would cause unnecessary load on the github server. This can however be solved by uploading the packaged variants of ugs (using: mvn package -pl ugs-platform/application -P create-autoupdate,create-dmg -DskipTests=true).

We have an artifactory server, I think it would be better to use that for hosting the binaries.

mike-pittelko:
I'm fine with that. Didn't know there was an artifactory instance, and don't have access to it that I'm aware of.  I was trying to provide -something- in the absence of anything - there haven't been any new binary nightlies available since December.

breiler:
@mike-pittelko I agree with Will, thank you for looking at this! And excuse me for sounding grumpy. =)

I'm also eager to get the new version out there as I have some new features I've been working on that I would like to share.

PR #1212, opened 2019-04-11
Feature/cli pendant

breiler:
This pull request will make it possible to start UGS with the web pendant in terminal mode. Also added a function for defining a workspace directory which basically is just useful from pendant api right now.

Added a virtual state for homing, this is required so that an external program can trigger that the homing cycle is finished by checkning the state transition from HOME to IDLE.

Moved the Keep Alive-function from the backend to the UI:s. Otherwise exceptions are thrown when using Java headless.


ugs-core/src/com/willwinder/universalgcodesender/uielements/panels/ConnectionSettingsPanel.java
```diff
@@ -142,5 +150,48 @@ protected void updateComponentsInternal(Settings s) {
...
+        panel.setLayout(new MigLayout("insets 0", "fill"));
+        panel.add(workspaceDirectory, "gapright 0");
+        panel.add(workspaceDirectoryBrowseButton);
+        workspaceDirectoryBrowseButton.setAction(createBrowseDirectoryAction());
+
+        add(new JLabel(Localization.getString("settings.workspaceDirectory")), "gapleft 56");
+        add(panel, "grow, wrap");
+    }
+
+    private AbstractAction createBrowseDirectoryAction() {
```
winder: You can put this in `SwingHelpers.java`

ugs-core/src/com/willwinder/universalgcodesender/utils/KeepAwakeUtils.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * A static util that will schedule a timer which will prevent the computer
+ * to go to sleep.
+ *
+ * @author wwinder
+ */
+public class KeepAwakeUtils {
```
winder: :+1:

PR #1219, opened 2019-04-26
add reload gcode button to quickly realod the  currently open gcode file

DanielWeigl:
Im not used to netbeans-platform, and there are some open points that might be an easy fix:
 - [fixed] somehow i changed the sort order of "File" and "Edit" menu bar (at least on my setup):    
image
- [fix] the enable logic is somehow not working completely. It works in the menu (the entry gets enabled if there is a gcode file loaded and it can be reloaded), but the enabled state does not get updated in the toolbar (only if i move the toolbar it gets reevaluated correctly), but the logs suggest that `setEnabled` is called correctly...


breiler:
Regarding the enable-logic, I think that this has to do with the property for enable isn't updated thus not firing an event which would update the button state and icon:
Line 233: https://zgrepcode.com/java/openjdk/10.0.2/java.desktop/javax/swing/abstractaction.java

Some of the actions have work arounds for this, example:
https://github.com/winder/Universal-G-Code-Sender/blob/master/ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/PauseAction.java#L68

It's kind of an hack, and should probably be fixed in a better way in all actions...

DanielWeigl:
Hi, thx for your reply, I have the same "hack" here: https://github.com/winder/Universal-G-Code-Sender/pull/1219/files#diff-221ad0c48dfe3f3d65b6abf984ed01c5R81 and also the log suggests that it gets called after a gcode file is loaded, but it does not change the state in the toolbar

breiler:
> Hi, thx for your reply, I have the same "hack" here:
> 
> [Universal-G-Code-Sender/ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/ReloadGcodeAction.java](https://github.com/winder/Universal-G-Code-Sender/blob/c9060d62e25a10b0116ac1aa173c98c540a68499/ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/ReloadGcodeAction.java#L81)
> 
> Line 81 in [c9060d6](/winder/Universal-G-Code-Sender/commit/c9060d62e25a10b0116ac1aa173c98c540a68499)
> 
>  if (cse.isFileChangeEvent()) { 
> and also the log suggests that it gets called after a gcode file is loaded, but it does not change the state in the toolbar

Oh, sorry about that! I was in a rush to answer without double checking your implementation. Wierd that it doesn't work. I can have a look at it tonight.

winder:
This looks right to me, try doing a clean build. Sometimes the xml files get stuck in a stale state and end up calling old versions of your code.

DanielWeigl:
Okay, found the hack to fix the hack ;)

The initial state of the button is disabled, but internally it is tracked as enabled, so the first setEnable(true) does not fire `firePropertyChange` in  `javax.swing.AbstractAction#setEnabled`
By setting it in the init of the action, it works

PR #1216, opened 2019-04-24
The jog action in visualizer now shows the position in preferred units.

breiler:
Fixes #921

PR #1220, opened 2019-04-26
fixes #1154 - fix the radius-calculation to respect the current selected working plane

### LINKED ISSUE #1154: Z Plane (YZ) Arc problem
## Problem description 
 Arc's in the YZ plane do not appear to be implemented correctly.  I need to translate an object from the XY to the YZ planes for an operation on my CNC machine.  I was successful in developing my G-code translation algorithm, as confirmed by proper operation of a different sender (OpenBuilds), but was never successful in getting UGS to properly visualize/send the translated G-code.

### Expected Behavior

Arcs in YZ plane behave the same as those in XY plane

### Actual Behavior

G0/G1 commands ported from XY to YZ planes are reproduced accurately, G2/G3 commands are not..

1. So far as I can tell, try any G2 or G3 command in a Z plane (I have focused on YZ, but believe it occurs in XZ as well).

2 .Here is an object G-code in XY that includes arcs:
[R_XY.txt](https://github.com/winder/Universal-G-Code-Sender/files/2620858/R_XY.txt)

Here is the UGS Visualizer in XY:
ugs xy screenshot

 3. Here is the same object G-code, translated to YZ:
 [R_YZ.txt](https://github.com/winder/Universal-G-Code-Sender/files/2620859/R_YZ.txt)

And the result in UGS Visualizer in YZ:
ugs yz screenshot

4. As a reference, I used a different sender (OpenBuilds), and the Z-plane operation worked fine on my CNC machine.

#### Version

UGS Platform 2.0.0 Stable

#### Operating system

Windows (both 7 and 10)

#### Platform

GRBL 1.1f, in a Workbee CNC machine with Laser attachment.  Translation is required to enable engraving on the side of a work piece.

DanielWeigl:
In `addArcPointSegment` the radius calculation was still hardcoded to XY-plane. If changed to respect the current active Plane, it now works

Now we also initialize the starting-Point with the correct Unit (it was not really triggering the bug, but it was throwing me off in the wrong direction, and might cause other bugs in the future) - It was `UNKNOWN`, but the end-coordinate was `MM` and i thought this is somehow leading to wrong conversations.


ugs-core/src/com/willwinder/universalgcodesender/gcode/GcodeParser.java
```diff
@@ -320,17 +319,19 @@ private static PointSegment addArcPointSegment(Position nextPoint, boolean clock
...
 
-        // Calculate radius if necessary.
+        // Calculate radius if necessary, according to the current G17/18/19 Plane
         if (Double.isNaN(radius)) {
+
             radius = Math.sqrt(
-                    Math.pow(state.currentPoint.x - center.x, 2.0)
-                            + Math.pow(state.currentPoint.y - center.y, 2.0));
+                    Math.pow(plane.axis0(state.currentPoint)  - plane.axis0(center), 2.0)
+                            + Math.pow(plane.axis1(state.currentPoint) - plane.axis1(center), 2.0));
```
winder: nice catch :+1:

DanielWeigl:
The failing test `com.willwinder.universalgcodesender.gcode.GcodeParserTest#autoLevelerProcessorSet` might also need a new `resources/gcode/circle_test.nc.processed` reference-file

grafik

ugs-core/src/com/willwinder/universalgcodesender/visualizer/GcodeViewParse.java
```diff
@@ -170,7 +172,18 @@ private static GcodeParser getParser(double arcSegmentLength) {
...
+                            start=new Position(
+                                    meta.point.point().x,
+                                    meta.point.point().y,
+                                    meta.point.point().z,
+                                    gp.getCurrentState().isMetric ? UnitUtils.Units.MM : UnitUtils.Units.INCH
+                                    );
+                        } else {
+                            // ...otherwise recycle the old instance and just update the x,y,z coords
+                            start.set(meta.point.point());
+                        }
```
winder: I thought the visualizer converts everything into MM, did you have a test case that this fixes?

DanielWeigl: Only via a manual debug test, i noticed that the start-point has no unit set:

grafik

winder:
@DanielWeigl that could be related to #1216

ugs-core/src/com/willwinder/universalgcodesender/visualizer/GcodeViewParse.java
DanielWeigl: And with the fix applied: 
grafik

But I think currently this does not really change anything, I just thought it would be more correct to have the correct units here

winder:
I believe I generated that file and manually observed that it looked right-ish, so just updating it is fine

DanielWeigl:
Ive updated the reference gcode file for the unit test and checked it against ncviewer.com, looks about right: 
grafik

DanielWeigl:
Ill implement some unit tests explicitly for the ArcExpansion later on to check some more "corner" cases

DanielWeigl:
Maybe OT (or not?): I just notice something strange with the streaming gcode output of the above test: https://pastebin.com/raw/BDxBizdA  

If i render this in NcViewer.com it has some strange jumps in the arc, is that expected or might this be another bug: 
grafik

DanielWeigl:
While playing with some testcases i noticed that a helix-arc is offset a slight bit after ArcExpansion: 
grafik

--> is fixed in cf9750be4c3bf85bd487a55ba59633e3ec4b13ba

(the error weights more the smaller the `numPoints/helixHeight` is)

winder:
I'm not sure what those lines you're seeing are. Is this visualizing the unexpanded, and expanded gcode at the same time? Did cf9750b fix that as well?

DanielWeigl:
Maybe I misunderstood what the streaming representation of the gcode is... Is it likely that ncviewer will not correctly unterstand it and maybe draw the original command and the expanded movements twice?

No, this fix does not change much in this regards, only the coords changed slightly.

Am 28. April 2019 00:18:17 MESZ schrieb Will Winder <notifications@github.com>:
>I'm not sure what those lines you're seeing are. Is this visualizing
>the unexpanded, and expanded gcode at the same time? Did cf9750b fix
>that as well?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/winder/Universal-G-Code-Sender/pull/1220#issuecomment-487324009

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

winder:
I didn't realize you were loading up the "streaming" file in there. Thats an internal format, basically just enough metadata to show the before, after and original line number. I'm surprised that ncviewer can display it at all!

These changes look ready to me, is there still WIP?

Thanks for fixing this bug!

DanielWeigl:
Ive added a bit more generic gcode-input-to-parsed-output comparison unit test "framework", maybe it can be used for other processors too and help to highlight unwanted side-effects of code changes. (I've called it "fixtures" - maybe in the CNC-context this might be a bit misleading?)

> These changes look ready to me, is there still WIP?
Okay

> Thanks for fixing this bug!
np, thx for working starting this great project

PR #1223, opened 2019-04-30
Refactor visualizer context - include option to jog or set offset 

DanielWeigl:
grafik

Move the "Jog to...." into a submenu and a more coordinates-contextual options
 - jog
 - set offset
 - copy

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/model/PartialPosition.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/model/PartialPosition.java
@@ -1,0 +1,144 @@
+package com.willwinder.universalgcodesender.model;
+
+
+import com.google.common.collect.ImmutableMap;
+import com.sun.tools.javac.util.Assert;
+
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a maybe partial coordinate (ie. only certain axis-values are known, eg. for moves where only some axis
+ * are changed)
+ */
+public class PartialPosition {
+    private final Double x;
+    private final Double y;
+    private final Double z;
+
+    public PartialPosition(Double x, Double y) {
+        this.x = x;
+        this.y = y;
+        this.z = null;
+    }
+
+    public PartialPosition(Double x, Double y, Double z) {
+        this.x = x;
+        this.y = y;
+        this.z = z;
+    }
+
+    // shortcut to builder (needed, because of final coords)
+    public static PartialPosition from(Axis axis, Double value) {
+        return new Builder().setValue(axis, value).build();
+    }
+
+    public boolean hasX() {
+        return x != null;
+    }
+
+    public boolean hasY() {
+        return y != null;
+    }
+
+    public boolean hasZ() {
+        return z != null;
+    }
+
+    public Double getX() {
+        Assert.checkNonNull(x, "Tried to get x-axis which is not set");
+        return x;
+    }
+
+    public Double getY() {
+        Assert.checkNonNull(y, "Tried to get y-axis which is not set");
+        return y;
+    }
+
+    public Double getZ() {
+        Assert.checkNonNull(z, "Tried to get y-axis which is not set");
+        return z;
+    }
+
+    public Map<Axis, Double> getAll() {
+        ImmutableMap.Builder<Axis, Double> allSetCoords = ImmutableMap.builder();
+        if (hasX()) {
+            allSetCoords.put(Axis.X, x);
+        }
+        if (hasY()) {
+            allSetCoords.put(Axis.Y, y);
+        }
+        if (hasZ()) {
+            allSetCoords.put(Axis.Z, z);
+        }
+        return allSetCoords.build();
+    }
+
+    public double getAxis(Axis axis) {
+        switch (axis) {
+            case X:
+                return getX();
+            case Y:
+                return getY();
+            case Z:
+                return getZ();
+            default:
+                return 0;
+        }
+    }
+
+    public static final class Builder {
+        private Double x = null;
+        private Double y = null;
+        private Double z = null;
+
+        public PartialPosition build() {
+            return new PartialPosition(x, y, z);
+        }
+
+        public Builder setValue(Axis axis, Double value) {
+            switch (axis) {
+                case X:
+                    this.x = value;
+                    break;
+                case Y:
+                    this.y = value;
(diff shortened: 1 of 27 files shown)


DanielWeigl:
How is the localization done? Do you use an external service or do you directly edit the `MessagesBundle_.....properties` files?

I have renamed and changed an entry in `MessagesBundle_en_US.properties`, so there are some stale entries in the other files - whats the best workflow here?

winder:
We're using POEditor to allow non-technical folks to easily contribute: http://winder.github.io/ugs_website/contributing/#translations

I have a script that I run periodically to synchronizes things using the POEditor API: [HERE](https://github.com/winder/Universal-G-Code-Sender/blob/master/scripts/update_languages.py)

I usually avoid renaming properties because it doesn't play very well with POEditor / my script, there are quite a few inconsistent properties at this point because of this.

DanielWeigl:
In this case I renamed it because the content also changed (no format strings anymore), so we don't get wrong content from different translations. Just need a simple way to remove the stale ones... will have a look at the editor

winder:
If the content changes, it is automatically marked as "fuzzy" in POEditor. But you would have to try replacing `%f` -> `%s`

DanielWeigl:
> I have a script that I run periodically to synchronizes things using the POEditor API: HERE

Okay, so ill edit them via POE editor once this branch is merged - for now EN is correct.

I have added a new abstraction (in https://github.com/winder/Universal-G-Code-Sender/pull/1223/commits/ce06240eb4c2609936f2e72dd09217894115a651 ) to represent one or more axis-values via a `PartialPosition`-class (which eg can format to X:10.0/Z5.0 -> `X10 Z5`). Is this refactor okay for you or do you already have something comparable in use (did not find anything)

If its okay for you, you can merge this PR

ugs-core/src/com/willwinder/universalgcodesender/model/PartialPosition.java
```diff
@@ -0,0 +1,149 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a maybe partial coordinate (ie. only certain axis-values are known, eg. for moves where only some axis
+ * are changed)
+ */
+public class PartialPosition {
```
breiler: I like this, nice work. 

Did you consider using Optionals which would allow us to scrap the has-methods. The get-methods would just return a `Optional.ofNullable(x)`.
I don't know if it gets any cleaner though.  To use one could do:
```
partialPosition.getX()
    .ifPresent(x -> sb.append("X").append(Utils.formatter.format(x)).append(" "));
```

winder:
One thing I would like to make sure of, is that we move towards consolidating how a position is stored. This fragments it slightly. The reason for this is to enable >3 axes in the future, I have some experiments related to that here: https://github.com/winder/Universal-G-Code-Sender/commits/abc-axis

One thing I did in that PR was remove `Point3d` from the code and replace them with `Position`. It actually sort of works: https://github.com/winder/Universal-G-Code-Sender/issues/1068#issuecomment-402120810

To that end, the idea of PartialPosition is great, I like how it enhances all the other APIs currently using Axis/double. But it also adds another place where x/y/z is hard coded. I think you could get the same flexibility by treating `PartialPosition` as an operation which can be applied to a `Postion`. I'm thinking the storage format would be something like the map you have in `getAll()`, and you would use the Builder exclusively to add the Axis/Position pairs. Then the controller could take those and do something like `position.apply(partial)`, or `position.applyOffset(partial)`.

Sorry to be difficult about this. What do you think?

DanielWeigl:
@winder - no, actually sounds good - and i also thought about it. One idea I had was to either 

- make `Position` a subclass of `PartialPosition` (then we can step-by-step look into functions that accept `Position`, if they would should use `PartialPosition`
- An interface `AxisPosition` and both `Position` and `PartialPosition` implement it and use this interface in most cases where we now use Position 

But both changes sounded like a bigger refactior and that they need their own branch ;)

ugs-core/src/com/willwinder/universalgcodesender/model/PartialPosition.java
DanielWeigl: The idea was that `PartialPosition` handles the String formatting... but for some reasons I put them into `GcodeUtils.formatPartialPosition(xyPosition)` -> ive now moved it into `PartialPosition` to make this more clear. Then it shouldnt be necessary (at least not that often) to individually check on `hasX()`...

DanielWeigl:
>I'm thinking the storage format would be something like the map you have in getAll()

I think we should have explicit implementations for Point3D (and maybe Point4D / Point5D?) so that we dont need to create  and hold an Map in memory for each point. And also maybe have a generic which internally uses the HashMap.

Could you try to rebase your `abc-axis` branch on master (or on this branch)? Then we can try this changes there.

DanielWeigl:
I just noticed that Jogging via context menue now includes a Z-coordinate, which is not expected - will also refactor to use `PartialPosition` for jogging and also move Units into `PartialPosition`

DanielWeigl:
Ive now also changed Jogging (via `JogService`) to use PartialPosition - now Z-coordinates are not anymore included if jogged via the ContextMenu.

winder:
I just finished merging master into the `abc-axis` branch

ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/actions/SetWorkingCoordinatesHereAction.java
```diff
@@ -0,0 +1,53 @@
...
+        if (position == null || !backend.isConnected()) {
+            setEnabled(false);
+        }
+    }
+
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        try {
+            backend.setWorkPosition(new PartialPosition(position.getX(), position.getY()));
```
winder: `PartialPosition.fromXY(position);`

winder:
This looks pretty much ready to get merged in, was there anything else you wanted to add? I think that if we ever add >3 axis it should be pretty straight forward to extend `PartialPosition` if needed.

DanielWeigl:
is okay for me

PR #1231, opened 2019-05-05
Point3D -> Position and start of ABC-Axis Visualizer

winder:
This is an incremental move towards supporting an arbitrary number of axes, mostly it is the backend support by removing vecmath Point3D from the backend. I believe this is safe to merge in, but need to re-review the work myself since it was written a while ago.

The biggest change is that the mixed Point3d/Position types were merged and replaced with the start of a type hierarchy where `Position` extends `CNCPoint`. @DanielWeigl made a suggestion about storage efficiency, and I was thinking about this as well. I think that the `Position` (or CNCPoint?) interface can serve as a starting point and there can be multiple concrete implementations which do not store the unused fields (i.e. a `PointXY` type could default to `0` if Z, A, B, or C is requested). Another idea was to have the underlying storage be an array, and concrete implementations could override the array index for accessing the data.

Curious about your thoughts @breiler  and @DanielWeigl


DanielWeigl:
Hi, sounds interesting, ive made a few lines of some scratch code (outside of the existing code, so I dont need to refactor everything, just to get some feeling for the idea) here: https://github.com/DanielWeigl/Universal-G-Code-Sender/blob/a8f9cbf5b19829a7c86c970980f2419c2ce9f38a/ugs-core/test/com/willwinder/universalgcodesender/model/test1/PartialPositionTest1.java#L10

Things ive tried:
- Have a generic base Interface `AxisPosition`which we can use, if we dont care which infos we have/need (eg to format a "G0/G1/..." move)
- Have  more specific and also memory optimized special implementations for 2D/3D points, and also allow functions to specify they need a 2D/3D point as a parameter and thus enforce this on a type level)
  - a 2D function can also accept a 3D point of course, but currently needs to cast it down, if it will stringify it.

Some points Im not sure:
- Should we strive for immutable position classes (would make threading easier, but might need a bigger refactor)
- Should every position have `Units` or would it make sense to internally always use `[MM]` and just convert if needed 
  - does Units.UNKNOWN make sense then?


This is just a quick and non-exhaustive test, maybe a good starting point for some further experiments (i dont have that much time per day to dive further into it...)

winder:
@breiler are you still ok with these changes? I'd like to merge them. Surprisingly the gigantic merge had very few conflicts and things still seem to work fine.

abc

breiler:
Sure, these are good changes. 

I remember trying this out and could not get the models I had with more than 3 axes to render properly. But this is a great start!

I started some work with the DRO, It will be nice to be able to extend it with extra coordinate fields.

PR #1243, opened 2019-06-07
Upgrading to NetBeans 11.0

breiler:
Attempt to upgrade netbeans to version 11.0. Has successfully managed to open netbeans with JDK8 and JDK12. Things seems to work fine, but might need additional testing before merging.


winder:
Wow!!!

winder:
I think this is on maven central now, so we can remove the `repositories` section of `ugs-platform/pom.xml`.

PR #1250, opened 2019-07-07
Error popups replaced with error messages in console

breiler:
When an error occurs in the controller an error popup is created. Multiple errors results in multiple popups which doesn't seem user friendly. Also when running a headless instance these popups are created.

image

I think it's better to log the errors in the console. In NetBeans notifications can also be created which wouldn't interfere with user interactions:
image

This pull request also replaces the code in MainWindow for the console with the same UI-component used in the platform edition.


winder:
Thanks, this looks great!

PR #1253, opened 2019-08-07
added GRBL v1.1  jog syntax support to jogMachineTo (G90) jog command

rhoffman12:
I'm working on my own joystick / joypad jogging plugin for UGS with GRBL, based on the GRBL team's [suggested implementation](https://github.com/gnea/grbl/wiki/Grbl-v1.1-Jogging#joystick-implementation). Adding support for $J jog commands to the jogMachineTo method (analogous to how it was done for jogMachine) lets such a plugin overcome the limited choices of jogging directions available in jogMachine, allowing for a much more satisfying / responsive joypad interface.


winder:
Thanks for the PR!

PR #1260, opened 2019-08-19
Preparing for new controller implementations

breiler:
I'm preparing for work on new controller implementations (Marlin, Smoothieware) and improving existing implementation for TinyG and g2core).

 - [x] Removed the command queue in the abstract controller which duplicates the queue in the current communicator implementation
 - [x] Made interfaces for the communicator and gcode streamer
 - [x] Simplified the registration of SerialCommunicatorListener. (I managed to forget to register the right type of listener for the third time, so I thought it was time to simplify this using only one type of listeners)
- [x] fixed bug when streaming + pausing + resuming a file when using check mode caused the file stream to remain paused.
- [x] It's now required to package the commands as GcodeCommand instead of using strings when sending commands through the communicator.
- [x] Fixed a couple of issues with the command queue count that made gui show negative command count and negative estimated time.


ugs-core/test/com/willwinder/universalgcodesender/BufferedCommunicatorTest.java
```diff
@@ -594,10 +594,5 @@ public boolean processedCommand(String response) {
         public boolean processedCommandIsError(String response) {
             return (response != null && response.startsWith("error"));
         }
-
-        @Override
-        public String getLineTerminator() {
-            return "\r\n";
```
winder: Where do these line terminator constants live now?

ugs-core/src/com/willwinder/universalgcodesender/utils/SimpleGcodeStreamReader.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * A simple Gcode stream for transmitting commands to a controller using string
+ * as an input.
+ *
+ * @author Joacim Breiler
+ */
+public class SimpleGcodeStreamReader implements IGcodeStreamReader {
```
winder: Great idea!!

ugs-core/src/com/willwinder/universalgcodesender/uielements/components/GcodeTable.java
```diff
@@ -22,58 +18,26 @@ This file is part of Universal Gcode Sender (UGS).
...
 /**
+ * Customized JTable with helpers to coordinate common operations.
  *
  * @author wwinder
  */
 public class GcodeTable extends JTable {
-    private static final Logger logger = Logger.getLogger(GcodeTable.class.getName());
-
-    private DefaultTableModel model = null;
+    private GCodeTableModel model = null;
```
winder: Nice cleanup. In the past this widget was one of the biggest performance bottlenecks of the classic GUI, it looks like some of these changes could have helped.

ugs-core/src/com/willwinder/universalgcodesender/types/GcodeCommand.java
```diff
@@ -27,28 +19,40 @@ This file is part of Universal Gcode Sender (UGS).
...
+    private static final AtomicInteger ID_GENERATOR = new AtomicInteger(0);
     private String command;
     private String originalCommand;
     private String response;
-    private String responseType;
     private Boolean sent = false;
-    private Boolean done = false;
     private Boolean isOk = false;
     private Boolean isError = false;
+    // TODO: Get rid of the commandNum member which is used in the GUI for indexing a table of these objects.
```
winder: I think that the command num is also used by gcode processors. For example the "arc expander" will split an arc into potentially hundreds of small line segments, each of which will retain the same command number. This way if a program were to fail, we could theoretically trace the program back to a line number in the file.

The gcode editor also used this to let you click on the arc command and highlight all of the line segments in the visualizer.

ugs-core/test/com/willwinder/universalgcodesender/BufferedCommunicatorTest.java
breiler: This method was never used and I was a bit surprised that it's hard coded in the BufferedCommunicator. If there is need for such a method I think it should exist in the Controller as the line terminator is protocol-specific?

ugs-core/src/com/willwinder/universalgcodesender/types/GcodeCommand.java
breiler: I think it's a good idea to still have the command number, I just moved the TODO-comment from the file header thinking that you might have a bigger plan. I'll remove the TODO.

ugs-core/test/com/willwinder/universalgcodesender/BufferedCommunicatorTest.java
winder: Interesting, it looks like this was deprecated last year by the `ResponseMessageHandler`. It is hard coded in subclasses of BufferedCommunicator, along with some other simple extensions to the communicator.

Things like pausing the stream if there is an error don't work as well if it has to wait for the events to be sent all the way back to the controller, there were other weird cases where the controller would be paused but the controller wasn't (and vice-versa). Probably fine to leave it this way, but it does make the "Communicator" abstraction less useful. i.e. an alternative `RxCommunicator` couldn't be dropped in.

Issue #1193, opened 2019-02-15
UGS locks up and continues jogging (video link in description)


407pilot:
## Problem description 
UGS locks up and continues jogging when using touchscreen.

### Steps to Reproduce the Problem
Tap jog in any direction 2+ times using touchscreen. Machine will initially obey the jog command, then will take off using smaller jogging steps. 

See video link at bottom to see machine and UGS behavior. Frankly, I don't really know how to describe what it's doing. 

#### Version
UGS Platform 2.0 - Nightly Build Dec 18, 2018

#### Operating system
Raspbian Stretch on Raspberry Pi 3b+

#### Platform
Inventables X Carve
X Controller
GRBL 1.1

#### Other
https://youtu.be/d0cewo06HfY

407pilot:
I shutdown and power cycled everything and now it won't reproduce. While this is definitely related to the touchscreen, I don't think it's solely a touchscreen problem since the jog distance changed and the program became unresponsive, even to regular mouse clicks. I'll keep playing around to see if I can get it to reproduce.

breiler:
This could have something to do with the continous jogging feature. If you find anything please let us know!

AbySet:
On windows when we click and hold at any direction, it does those small steps, that's normal behaviour, its a feature in the program, see your OS touch screen config, maybe it's set to switch to hold when you double tap, i personaly don't know how your OS work, but you can look for eventual config that make double tap do a holding tap.

pmckeon:
I've had this happen to me a few times on Windows when repeatedly clicking a jog button in any direction, maybe it sometimes fails to detect the mouse up/button up event and will jog indefinitely until stopped.

AbySet:
I'm just a user like you @pmckeon, allow me to recomand you to make a new post about your issue, you are using windows (In this post @407pilot is using a Raspbian Stretch on Raspberry Pi 3b+) the problem may have different reason and i believe you will have better help with a new post.

PR #1293, opened 2019-10-24
Fixed problem with continuous jogging hanging

### LINKED ISSUE #1126: Jog -X-Y runs away
Using the jog controller, when you hold the -X-Y control for several seconds -X-Y direction locks on as it should.  In major X,Y,Z directions clicking the field stops motion.  However -X-Y motion can not be stopped, the machine runs away.
### Expected Behavior
Clicking the -X-Y feild should stop motion.
### Actual Behavior
<the machine runs away

 1.  Open the program
2. connect to machine (milling machine in my case)
3.  Open Jog Controller
4.  click and hold -X-Y for about 3 seconds
5.  -X-Y travel should begin and continue after releasing mouse click.
6.  Attempt to stop motion by clicking -X-Y field  (fail)

#### Version
Universal G-Code Platform Nightly 9/9/18. Running on Windows 10 ver 10 running x86 CP1252 en US

breiler:
This is a potential fix for the problems with continuous jogging hanging. I've moved the code for continuous jogging to its own handler which is responsible for checking if the command has entered GRBLs planner before issuing a cancel jog. It's not pretty but it looks like it's working.

Fixes #1126 and #1193

Thanks to @AlexHolden, @bdurbrow and @zeevy for helping out!

### DIFF
--- a/ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/ContinuousJogHandler.java
+++ b/ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/ContinuousJogHandler.java
@@ -62,30 +62,10 @@ public class ContinuousJogHandler implements ControllerListener {
         if (isRunning) {
             return;
         }
-
+        
         isRunning = true;
         isWaitingForCommandComplete = false;
-
-        final double stepSize = calculateStepSize();
-        ThreadHelper.invokeLater(() -> {
-            try {
-                while (isRunning) {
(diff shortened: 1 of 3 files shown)


lorenzo535:
I am still regularly have the jog mode remain stuck and the X or y axis run away. No commands (stop, soft reset) are accepted by the ugs GUI. Only course if action is the emergency reset button or power off the machine.
The bug happens on a regular basis in my case. I'm using a Xbox joypad. Bug happens regularly on a grbl 1.1 MAKERBASE MKS dlc32

breiler:
@lorenzo535 please create a new issue. This code change has nothing to do with jogging with gamepads.

PR #1314, opened 2019-11-29
Some refactoring of the ControllerStatus.

KawateiHikaru:
While building my CNC needle cutter for DTFB airplanes (#FliteTest, #ERC #TimSav), I had to build an load a GRBL controller. Since I don't know much about GRBL or the communication between the host and the controller (I know some 3d printing G-codes but that's it). I figured I will browse through the code to see how the communications are done. I see that some of the codes for handling communication are a bit older and with its doing quite a bit of work to parse the inputs. I figure I will refactor this. The refactoring is not complete, the next piece I would like to tackle is the V1Format parsing and the overriding logic. Please let me know if you have any feedback. Thanks.


KawateiHikaru:
^ lol the good old, you touched the file you introduced the issue. ;p

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerStatusBuilder.java
```diff
@@ -107,12 +134,84 @@ public ControllerStatusBuilder setPins(ControllerStatus.EnabledPins pins) {
...
     }
 
-    public ControllerStatusBuilder setStates(ControllerStatus.AccessoryStates states) {
-        this.states = states;
+    public ControllerStatusBuilder setAccessoryStates(ControllerStatus.AccessoryStates states) {
+        this.accessoryStates = states;
         return this;
     }
 
+    public static ControllerState getControllerStateFromStateString(String stateString) {
```
breiler: The method for creating a controller state from string is hardware dependant and should not exist in the generic ControllerStatus.

I was hoping that the state string would be removed in favour of the more static enum types.

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerStatusBuilder.java
```diff
@@ -107,12 +134,84 @@ public ControllerStatusBuilder setPins(ControllerStatus.EnabledPins pins) {
...
+        else {
+            if (lastStatus != null ) {
+                overrides = lastStatus.getOverrides();
+                pins = lastStatus.getEnabledPins();
+                accessoryStates = lastStatus.getAccessoryStates();
+            }
+        }
+
+        if (!stateOverWritten)
+            state = getControllerStateFromStateString(stateString);
```
breiler: This will become a problem if I set the state and it gets overwritten by the state string.

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerStatusBuilder.java
KawateiHikaru: I see, let me rework this :)

ugs-core/src/com/willwinder/universalgcodesender/listeners/ControllerStatusBuilder.java
KawateiHikaru: I think my variable naming isn't the best, this is to prevent your state to be overwritten with state String.

PR #1321, opened 2019-12-10
Update update_languages.py

Cehos:


winder:
@Cehos I wrapped up the missing pieces, you should be able to select Czech in the next build, thanks for the contribution!

Issue #1151, opened 2018-11-18
UGS Platform crashes on start on ubuntu 18.10 with java 11


giasone70:
## Problem description 
Starting the program it crashes after displaying the nightly build warning.
Looking at the output in the console window you get this:

> WARNING: An illegal reflective access operation has occurred
> WARNING: Illegal reflective access by org.netbeans.ProxyURLStreamHandlerFactory (file:/home/lab/Programmi/ugsplatform/platform/lib/boot.jar) to field java.net.URL.handler
> WARNING: Please consider reporting this to the maintainers of org.netbeans.ProxyURLStreamHandlerFactory
> WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
> WARNING: All illegal access operations will be denied in a future release
> Initializing...
> Loading LocalizingService...
> Loading JogService...
> Loading ActionService...
> Loading MacroService...
> Loading SendStatusLineService...
> Loading SettingsChangedNotificationService...
> Loading WindowTitleUpdaterService...
> Services loaded!
> Setting UGP version title.
> Initializing firmware... ...
> /resources/firmware_config
> /resources/firmware_config/test_2.json
> /resources/firmware_config/smoothieboard.json
> /resources/firmware_config/grbl.json
> /resources/firmware_config/test_1.json
> /resources/firmware_config/tinyg.json
> Inconsistency detected by ld.so: dl-lookup.c: 112: check_match: Assertion `version->filename == NULL || ! _dl_name_match_p (version->filename, map)' failed!
 

### Expected Behavior
Normal program start
### Actual Behavior
Crash

### Steps to Reproduce the Problem
<!-- Example:
  1. Open the program
  
  1. Program crashes

#### Version

  UGS Platform 2.0 - Nightly Build Nov 18, 2018

#### Operating system

  Linux Ubuntu 18.10

carneeki:
I also have problems with Java 11 (but not the same one) on Arch Linux (rolling release, well maintained & updated). Sometimes I can compile UGS, sometimes I have to roll back to Java 8, and I think mostly it's to do with NetBeans 8.x as a dependency. I suspect that there's a plan to move UGS to a more recent NetBeans version in the future, but maybe Will is just short for time.

Recommendation: try rolling back to Java 8. Here's a couple of ways for different distros, including this as it may prove helpful to others for other distros too.

| Distro | Command | Notes & Link |
| ----- | ----- | ----- |
[...]

PR #1334, opened 2019-12-30
Upgrade libraries.

winder:
I tried upgrading libraries in case there was some Java8 -> Java9 issue which might have been breaking the Java13 build. Turned out they were not a problem.

Any reason not to update these anyway?


breiler:
Looks good to me. I never got Netbeans 11.2 to start so I didn't bother  upgrading. But if this works, go for it! :smile:

breiler:
I've got this running on:
* MacOSX and Java 8
* Windows 10 and Java 13
* Ubuntu and Java 8 (could not get it to run with java 11, 12 or 13  where I get the error `Inconsistency detected by ld.so: dl-lookup.c: 112: check_match: 'Assertion version->filename == NULL || ! _dl_name_match_p (version->filename, map)' failed!`)

winder:
Merging this in anyway, looks like the ubuntu issue you noticed isn't new: #1151

PR #1336, opened 2020-01-02
Updated to OpenJDK 13

breiler:
Tried running on Mac, Windows and Linux using the variants using a bundled Java which worked on all platforms with Java 13.

It's maybe time to add download links?


winder:
Great, thanks!

PR #1339, opened 2020-01-04
Outline g-code program

breiler:
Added a feature that will move around the loaded model with rapid movement. The intention is to make sure the placement of the material will contain all carving motions.

The outline is constructed using a convex hull algorithm. Only carving motions (not rapid movements) are included.

The action is available in the context menu in the visualizer
outline

There is also an action that can be added in the main toolbar, but for now I just want to make sure it works before adding it:
Screenshot 2020-01-04 07 41 22

Do you have other suggestions for the action name and/or the icon please let me know!


khturnings:
This is great - something I have been after for a while. Could this be a dual action?
I would also like to know the center of the job - this will be handy when setting up Laser jobs.
So maybe left click for outline and right click for center or another icon for center.
Just a thought. 
Thanks for all the hard work you are all putting into this.

ugs-core/src/com/willwinder/universalgcodesender/utils/MathUtils.java
```diff
@@ -0,0 +1,123 @@
...
+     * To find orientation of ordered triplet (p1, p2, p3).
+     * https://www.geeksforgeeks.org/orientation-3-ordered-points/
+     *
+     * @param p1 a point in a polygon
+     * @param p2 a point in a polygon
+     * @param p3 a point in a polygon
+     * @return The function returns if the sequence of the points are COLINEAR, CLOCKWISE or COUNTER_CLOCKWISE
+     */
+    public static int orientation(PartialPosition p1, PartialPosition p2, PartialPosition p3) {
+        double value = (p2.getY() - p1.getY()) * (p3.getX() - p2.getX()) -
```
winder: Does this ever need to take the arc plane mode into consideration? Not having it be XY would by unusual but maybe there are cases where it matters?

ugs-core/src/com/willwinder/universalgcodesender/utils/MathUtils.java
```diff
@@ -0,0 +1,123 @@
...
+        return (value > 0) ? CLOCKWISE : COUNTER_CLOCKWISE;
+    }
+
+    /**
+     * Returns a list of points making a convex hull in the X/Y plane of a list of points.
+     *
+     * @param points a list of points to build a convex hull for
+     * @return an ordered list of points creating a hull for all points
+     */
+    public static List<PartialPosition> generateConvexHull(List<PartialPosition> points) {
```
winder: The communicator is probably a good test, since it is pretty big. Does this algorithm take a while to run?

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/OutlineAction.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private List<LineSegment> parseGcodeLinesFromFile(File gcodeFile) throws IOException, GcodeParserException {
+        List<LineSegment> result;
+
+        GcodeViewParse gcvp = new GcodeViewParse();
+        try (IGcodeStreamReader gsr = new GcodeStreamReader(gcodeFile)) {
+            result = gcvp.toObjFromReader(gsr, ARC_SEGMENT_LENGTH);
+        } catch (GcodeStreamReader.NotGcodeStreamFile e) {
+            List<String> linesInFile = VisualizerUtils.readFiletoArrayList(gcodeFile.getAbsolutePath());
+            result = gcvp.toObjRedux(linesInFile, ARC_SEGMENT_LENGTH);
```
winder: Maybe the back end shouldn't generate a file event until the initial parsing to avoid this edge case.

ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/actions/SetWorkingCoordinatesHereAction.java
```diff
@@ -27,7 +27,9 @@ This file is part of Universal Gcode Sender (UGS).
 import java.awt.event.ActionEvent;
 
 /**
- * An action that will jog to a coordinate
+ * An action for setting the work position for X/Y axis to given position
+ *
+ * @author Daniel Weigl
```
winder: Dan?

ugs-core/src/com/willwinder/universalgcodesender/utils/MathUtils.java
breiler: It takes about 430ms to load the model using the preprocessor and about 60ms to run the convex hull algorithm (at least on my computer).

I've added a progress dialog that will display that it's working with generating the gcode. This dialog should be usable when loading a new file as well. I've also added a minimum time that it's visible so that the user gets the time to read it.

ugs-core/src/com/willwinder/universalgcodesender/utils/MathUtils.java
breiler: No it doesn't. I'm not exactly sure how this is supposed to work. Do you have any ideas?

ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/OutlineAction.java
breiler: This code is duplicated in several places and should probably be extracted in some way. Is it ok if leave this as it is until we have the time to fix this?

winder: Yeah, just thinking out loud

Issue #1344, opened 2020-01-12
DRO changes to WCS offset ignore differences between DRO and machine units


dvsingletary:
## Problem description 
If Jog/DRO are showing units in inches, but machine state is in MM (G21), typing a raw value into any of the WCS values in the DRO will be sent to the controller without being converted into the units that the controller is currently using.  Same applies vice-versa (bug exists if the DRO/Jog units are different than the controller units)

### Expected Behavior
Typing in a value into the DRO should be interpreted in whatever units are active, and if this doesn't match the controller units, it should be converted to controller units when the G10 L20 command is issued.

### Actual Behavior
G10 L20 is sent to the controller using whatever value was typed into the DRO without conversion if needed.

### Steps to Reproduce the Problem
1) Set DRO/Jog units to Inches (using the inch/mm button in the Jog controls)
2) Set controller state to mm (issue G21 command in console)
3) click on any of the WCS coordinate values in the DRO to change it
4) enter a non-zero value
5) observe that the value is sent (via G10 L20) to the controller without converting into the controller's units
6) DRO shows a different value than what was entered, because it's showing the actual coordinate that the controller learned.

I'll take a crack at fixing this.

breiler:
I've noticed this a couple of times but wasn't able to reproduce it reliably. This is great if you manage to fix it.

dvsingletary:
I saw that there have been some changes to this code since I filed this issue, and I'm no longer to reproduce it (ie bug is fixed).

PR #1346, opened 2020-01-13
fix for issue #1344

dvsingletary:
The fix for issue #1344 was to ensure that the AbstractController.setWorkPosition() function was always passed a PartialPosition that had the units specified, and to convert the units to match the gcode state before sending to the controller.  Once this started working, I noticed that when the DRO was in mm, but the Gcode state was Inches, that we ran into round-off errors because PartialPosition was always truncating off to 3 decimal places regardless of the units, so I put in a fix for that, and as a result had to modify some of the test code due to the additional precision.  Now, regardless of what the Jog/DRO unit state is and the Gcode unit state, when you enter a position into an axis in the DRO it will always convert to the proper units for the G10 L20 command sent to the controller.  Some modification was needed in the mathematical expression evaluation code also to ensure the proper units were being used.  I tested this in each possible permutation of jog units and gcode units and it all seems to work - a little more involved than I first thought.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
@@ -169,7 +169,7 @@ public class GrblUtils {
     /**
      * Generate a command to set the work coordinate position for multiple axis.
      *
-     * @param offsets the new work position to use (one ore more axis)
+     * @param offsets the new work position to use (one ore more axis) (Note: must already be converted to gcode state units)
      * @param grblVersion the GRBL version
      * @param grblVersionLetter the GRBL build version
      * @return a string with the gcode command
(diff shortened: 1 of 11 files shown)


ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
```diff
@@ -169,7 +169,7 @@ protected static String getResetCoordToZeroCommand(final Axis axis, final double
     /**
      * Generate a command to set the work coordinate position for multiple axis.
      *
-     * @param offsets the new work position to use (one ore more axis)
+     * @param offsets the new work position to use (one ore more axis) (Note: must already be converted to gcode state units)
```
breiler: I'm thinking that the offsets should contain the units and that this could be handled internally in the controller implementation. But this works for now.

breiler: Or did you mean that the position needs an known unit for this to work?

dvsingletary: When adjusting the work offset via the DRO, it was being passed to the AbstractControler derived class as a PartialPosition without units set. My fix involves making sure the dro/jog units are set in that PartialPosition, and then I also modified both GRBL and tinyG controllers so they require the PartialPosition to have the units set, and they convert the units to the gcode state units.

PR #1369, opened 2020-03-19
Smoothieware support

breiler:
A first version with support for smoothieware.


ugs-core/src/com/willwinder/universalgcodesender/SmoothieCommunicator.java
```diff
@@ -19,9 +19,33 @@ This file is part of Universal Gcode Sender (UGS).
...
+
+    @Override
+    protected boolean processedCommandIsError(String response) {
+        return GrblUtils.isErrorResponse(response);
+    }
+
+    @Override
+    public boolean allowMoreCommands() {
+        // Force single step
+        return this.activeCommandList.isEmpty();
```
winder: You should be able to override `this.singleStepModeEnabled` (or override the setter to ignore anything that disables it) instead of making `activeCommandList` protected.

breiler:
> Cool, nice work. Out of curiosity, what were the main ways that smoothie/grbl were different? The code looks very similar.

I began subclassing the GRBL-implementation, but small subtle changes in the their protocols made it difficult... Now this was a couple of months ago but I remember that I ended up overriding most methods.

But there's a couple of things I've been struggling with:
* GRBL has a clear versioning which makes it easy to switch between protocol differences (or adding removing capabilities). In Smoothieware it's a build hash and a build date. This is no longer an excuse to have the code separated though, as I totally dropped the support for the legacy protocol for Smoothieware.
* Statuses and responses from the controller isn't always single line or end with a ok/error line.

I've fixed most of the large chunks of duplication, so it's at least better...

PR #1373, opened 2020-03-25
Joystick integration

breiler:
Added a plugin that will read joystick/gamepad values and send jog commands.

I've hard coded the controller buttons and axises accordingly. Future improvements could be to make the mappings configurable:
* Left axis is mapped to X/Y
* Left direction pad is mapped to X/Y
* Right axis is mapped to Z
* Button Y/A is mapped to Z
* Left/Right bumper button is mapped to multiply/divide feed speed
* Start button is mapped to start sending
* Select button is mapped to stop sending

One big change is within the jog service that no longer use integers for the direction. Instead it will use a float value that represents a percentage of the step size to be moved.

Once merged I'll write up some usage instructions.


ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
```diff
@@ -234,8 +234,8 @@ protected void softReset() throws Exception {
     }
 
     @Override
-    public void jogMachine(int dirX, int dirY, int dirZ, double stepSize,
-            double feedRate, UnitUtils.Units units) throws Exception {
+    public void jogMachine(float dirX, float dirY, float dirZ, double stepSize,
+                           double feedRate, UnitUtils.Units units) throws Exception {
```
winder: This is kind of cool, but having the direction + step size combination is a little confusing. It looks like you got it right when generating gcode (multiplying the two together).

Is this function still easy to work with? Would we lose anything by changing the method to `jogMachine(float distanceX, float distanceY, float distanceZ, double feedRate, UnitUtils.Units units)`?

Maybe from the joysticks perspective it's natural to have the direction be a ratio.

ugs-core/src/com/willwinder/universalgcodesender/utils/ContinuousJogWorker.java
```diff
@@ -26,63 +26,101 @@ This file is part of Universal Gcode Sender (UGS).
...
+     * <p>
+     * The step size is given using the current units of {@link Settings#getJogFeedRate()}.
+     *
+     * @return the step size in the units of {@link Settings#getJogFeedRate()}
+     */
+    private double calculateStepSize() {
+        double maxFeedRate = jogService.getFeedRate();
+
+        // Calculate how long we should be able to move at the given interval and add 20%
+        return (maxFeedRate / 60) * (JOG_COMMAND_INTERVAL / 1000.0) * 1.2;
```
winder: Should 60 be 60.0 to avoid rounding?

ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
breiler: I had to make a choice about a year ago to either create a completely new function for jogging all axises or to copy the direction concept when implementing the jog controller to keep things consistent.

A change like this will ripple all the way up to the jog service, but in the end it would probably become a cleaner solution. I can give it a try!

breiler:
Kaboom!

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/service/JoystickServiceImpl.java
```diff
@@ -0,0 +1,156 @@
...
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@ServiceProvider(service = JoystickService.class)
```
winder: I'm not sure you need this abstraction. The ServiceProvider interface is mainly for communication across modules. Like if you had the JoystickClass interface defined here and the implementation in a separate module.

ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
winder: It's up to you, this would definitely cause a big ripple.

winder: haha, too late on this comment. I think this will be nice! The function is definitely more intuitive this way.

breiler: Things got a lot simpler using distances instead of directional flags. I prefer things this way. =)

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/JoystickOptionsPanel.java
```diff
@@ -0,0 +1,78 @@
...
+
+public class JoystickOptionsPanel extends AbstractOptionsPanel implements JoystickServiceListener {
+
+    private final JoystickServiceImpl joystickService;
+    private JPanel panel;
+    private Checkbox activeCheckbox;
+
+    JoystickOptionsPanel(JoystickOptionsPanelController controller) {
+        super(controller);
+        joystickService = CentralLookup.getDefault().lookup(JoystickServiceImpl.class);
```
winder:
```suggestion
        joystickService = CentralLookup.getDefault().lookup(JoystickService.class);
```

breiler:
Thanks for your help!

I've tested it as best as I could virtually with GRBL and g2core and with a GRBL based CNC.

Issue #1352, opened 2020-01-28
Universal G-Code Sender running behind cnc machine


TheNetStriker:
I'm running UGS platform (I tried stable and nightly, but no difference there) on a Raspberry Pi 4 with a Sainsmart 3018 pro. When I run a bigger gcode file I have the problem that the Visualizer and the console output get's behind the actual movements of the CNC machine. When the machine finishes the Visualizer and the console is still running and I cannot abort the process and have to wait until the output in the console finished. (This can take several minutes) The only workarround I found so far is to close the Visualizer before starting the gcode file.

@KEITHABEAR also reported this issue in the grbl project: https://github.com/gnea/grbl/issues/789

Is this a known problem or is there a workarround for this besides simply closing the Visualizer?

breiler:
Yes this is a known issue. Would you mind sharing your gcode file for us to test with?

TheNetStriker:
@breiler Here is a link to the gcode files: https://www.dropbox.com/sh/jd5mvldl9bcdlyg/AADFC6u4TinjOLLcxEjf1I-ha?dl=0

The first file is small, the problem there is not noticable, but on the second and third file I had the problem.

KEITHABEAR:
Does any know how to fix this it started doing it again

breiler:
I've made some changes in the visualizer that might fix this. I wasn't able to access your gcode file so I couldn't test. I'm closing this, please reopen if the problem still exist.

Issue #1278, opened 2019-10-04
Console lags output


BallscrewBob:
Product Version: Universal Gcode Platform 201609300101
Java: 1.8.0_181; Java HotSpot(TM) 64-Bit Server VM 25.181-b13
Runtime: Java(TM) SE Runtime Environment 1.8.0_181-b13

Win 7 x64

Can also reproduce on latest version but one above is more stable here.
**NOTE the visualiser was OFF !** for the whole run.

Job has finished and the machine is at "safe Z" but not ZERO.
The console is still running and preventing other functions from working properly.
The file attached (and many others)  will eventually bring up the box  "Timeout waiting for GRBL to idle during cancel. cancel incomplete" I hit "OK" and can hit the stop button again but still zero response. Only way out is to close UGS and then re-find Z zero again for the job if I need to add another cut or open a second file for further processing on the job.

[130 dia 2mm ball FINE.zip](https://github.com/winder/Universal-G-Code-Sender/files/3690828/130.dia.2mm.ball.FINE.zip)

Found the log for this issue too but not sure it is going to be much help to you.
[messages.log.zip](https://github.com/winder/Universal-G-Code-Sender/files/3690851/messages.log.zip)

BallscrewBob:
Can also reproduce on this version.

Product Version: Universal Gcode Platform 20190423
Java: 1.8.0_181; Java HotSpot(TM) 64-Bit Server VM 25.181-b13
Runtime: Java(TM) SE Runtime Environment 1.8.0_181-b13
System: Windows 7 version 6.1 running on amd64; Cp1252; en_US (ugsplatform)
User directory: D:\Users\Bob\AppData\Roaming\.ugsplatform\2.0-SNAPSHOT\dev
Cache directory: D:\Users\Bob\AppData\Roaming\.ugsplatform\2.0-SNAPSHOT\dev\var\cache

rlwoodjr:
We are experiencing the same issue.

My understanding from our testing is that in a large file run at a "fast" feed rate causes a lag between the GUI and the serial communications. Therefore the GUI is not live. It lags behind. When we finish a large file (spindle returns home and the project is complete), the UGS screen still has 30% to 50% of lines to process. I actually let it continue to run a few times and it does finally finish up and give you back the GUI. This is true with 16Gb dual processor or 2Gb mini computer.

We started the same file on 2 different CNC's on at 100% and the other at 200%.   The computers were identical. 

[...]

Issue #766, opened 2017-09-17
Sender stop working 


inFamousMcGrath:
Hi, i'm using last nighty build of UGS Platform, and i have Arduino Mega 2560 with GRBL 1.1f.
I'm working with a large G-code files(aroud 1-2 million rows), which have onli G1, G0 and M3/M5 commends. So here is a problem - sender randomly stops to send a rows, without any messages. It shows only on big Gcode files (current file size is 25 MB). GRBL status in verbose is Idle, spindle don't disable, so it just stay on last XYZ position with enabled spindle)). 
p.s. Sorry for bad English.

inFamousMcGrath:
Update - it happened only on notebook with 1 GB ram and Atom n450. i7 6700 +8Gb ram works fine.

winder:
@inFamousMcGrath try closing the console window, if that doesn't work also close the visualizer. If one or both of those fixes it I can investigate why they're leaking memory over time.

inFamousMcGrath:
I'll try, but there more bad news by this hour - it's stopped even on i7 6700 - after 6 hours, but stopped. I cant stop sending, it dont react to my actions, only restart helps. There 111000 rows, that less that previous file, which was around 1000000 rows. Visualizer disabled, but console not, i'll try to disable it.
![default](https://user-images.githubusercontent.com/31566725/30563609-1034f948-9ccb-11e7-92b7-30edb7bc6d87.JPG)

vMeph:
that pendant sender have alot bugs, 
wish someone would fix,would be a great app for the smart phone

winder:
The pendant is another variable I wasn't even considering. When I wrap up with the features I'm working on now I'll try profiling the code to find where the memory is leaking. Until then if you leave the pendant disabled along with the console and visualizer hopefully the problem goes away. If it doesn't there is a much bigger problem than I realized.

inFamousMcGrath:
So, turning off the console doesn't helps me. It's strange, it happens only at night time, when i left my work))
Well, i have some ideas, what it will be:
Sender does not receive exact number of "ok" answers from GRBL (or it's an UGS bug), which need to continue sending. If there no error messages from UGS - i thought it really main option. I reccomend to synchronize UGS <ok> counter with verbose - GRBL 1.1 response number of queued motion anb avaliable buffer. Looks like Candle use right this method with GRBL 1.1.

Siand:
[...]

Issue #655, opened 2017-05-10
ugsplatform lags and freezes after stopping/pausing job


Harvie:
I've been using ugs platform to cut relatively big job (21MB g-code) and everything was cut perfectly. Job was finished in two hours. But when it stopped the ugs was still showing like 80% (10 hours approximated) remaining. And these coordinates followed by ok continued to run in console window:

![image](https://cloud.githubusercontent.com/assets/276504/25917953/0b2e6118-35ca-11e7-9859-2ea374b3642c.png)

The same happend when i tried to stop/pause the job in the middle. ugs was frozen except for showing coordinates that were not received by grbl. This even happened when i pressed reset button on grbl. New g commands still kept showing even when machine was already stoped.

winder:
What version of GRBL are you using? Could you share the gcode file?

Harvie:
grbl v1.1e, stopping mid job works perfectly with smaller files.
This is the g-code: [holes.txt](https://github.com/winder/Universal-G-Code-Sender/files/1034217/holes.txt)

winder:
@Harvie thanks for the file. After loading the file and running it for a bit I think this is most likely a  performance issue. The way the arcs are made using hundreds of short line segments means there is a lot of work to do for processing GRBL responses and notifying the GUI.

I was still able to pause the job, but sometimes I had to click it several times before it responded, probably some sort of race condition caused by UGS processing so many events.

Now the strangest part... you mentioned that the job finished but you still saw the gcode/ok messages coming through the console AND the machine location being updated?

Harvie:
Yes, IIRC the stop/pause button was greyed out after first click, but everything kept running. I think there should be something that will synchronize threads, so nothing will send more g-code commands to machine as soon as button was clicked. If there's some arc processing on background it can finish separately after sending is terminated.

BusbyGT:
I'm currently experiencing this same thing. The GUI worked for the first hour of the carve and then I got an update maybe an hour later that increments the elapsed time by like 20 minutes. I did a feed hold on my controller and let it sit idle for 1.5 hours and when I came back, still no GUI response until about 30 mins later and it catches up to the point where I did the feed hold. The job is still running with no issues, I can update when it's done cutting. It's an 11MB Gcode file that I'm streaming for a 3D finishing toolpath (629,000 lines).

GRBL v1.1f, Xcontroller, RaspberryPi 3, the July 23 nightly semi-stable build.

Harvie:
[...]

PR #1375, opened 2020-03-27
Runs the vertex updates in its own thread but only allows one update at the time

breiler:
Runs the vertex updates in its own thread but only allows one update at the time. This should fix some of the performance issues with large files with many small movements.

One way to confirm this is to use Bob's test file in #1278, my computer chokes after a couple of seconds. After this fix it works at least on my computer. =)

I think it also solves but have no way of confirming it: #1352, #655, #766,

### DIFF
--- a/ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/renderables/GcodeModel.java
+++ b/ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/renderables/GcodeModel.java
@@ -55,7 +55,7 @@ import static com.willwinder.ugs.nbm.visualizer.options.VisualizerOptions.*;
 public class GcodeModel extends Renderable {
     private static final Logger logger = Logger.getLogger(GcodeModel.class.getName());
 
-    private boolean colorArrayDirty, vertexArrayDirty;
+    private boolean colorArrayDirty, vertexArrayDirty, vertexBufferDirty;
 
     // Gcode file data
     private String gcodeFile = null;
@@ -94,7 +94,7 @@ public class GcodeModel extends Renderable {
         arcColor = vo.getOptionForKey(VISUALIZER_OPTION_ARC).value;
         plungeColor = vo.getOptionForKey(VISUALIZER_OPTION_PLUNGE).value;
         completedColor = vo.getOptionForKey(VISUALIZER_OPTION_COMPLETE).value;
-        updateVertexBuffers();
+        vertexBufferDirty = true;
         colorArrayDirty = true;
     }
 
@@ -120,7 +120,8 @@ public class GcodeModel extends Renderable {
      */
     public void setCurrentCommandNumber(int num) {
         currentCommandNumber = num;
-        vertexArrayDirty = true;
+        vertexBufferDirty = true;
+        colorArrayDirty = true;
     }
 
     public List<LineSegment> getLineList() {
@@ -162,6 +163,10 @@ public class GcodeModel extends Renderable {
                 && gl.isFunctionAvailable( "glDeleteBuffers" ) ) {
             
             // Initialize OpenGL arrays if required.
+            if (this.vertexBufferDirty) {
+                updateVertexBuffers();
+                this.vertexBufferDirty = false;
+            }
             if (this.colorArrayDirty) {
                 this.updateGLColorArray(drawable);
                 this.colorArrayDirty = false;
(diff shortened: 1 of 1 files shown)


ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/renderables/GcodeModel.java
```diff
@@ -287,62 +288,73 @@ private boolean generateObject()
...
-                    color = arcColor;
-                } else if (ls.isFastTraverse()) {
-                    color = rapidColor;
-                } else if (ls.isZMovement()) {
-                    color = plungeColor;
-                } else {
-                    color = linearColor;
-                }
+        // Prevent concurrent updates which will queue events
+        if(isUpdatingVertexBuffers) {
```
winder: There's technically a race condition here, I think this function needs to be synchronized, and set the isUpdatingVertexBiffers flag outside the invokeLater.

ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/renderables/GcodeModel.java
```diff
@@ -287,62 +288,73 @@ private boolean generateObject()
...
+            return;
+        }
 
-                // Override color if it is cutoff
-                if (ls.getLineNumber() < this.currentCommandNumber) {
-                    color = completedColor;
+        // Start the update in it's own thread so we aren't blocking anything.
+        ThreadHelper.invokeLater(()-> {
+            isUpdatingVertexBuffers = true;
+            if (this.isDrawable) {
```
winder: Wrap this in a try/catch to ensure we set isUpdating back to false

winder:
Great find! I guess that with this change the visualizer will skip frames instead of hanging the entire event stream?

breiler:
It was apparently a bad idea to update the buffers in it's own thread. When testing further it crashed the JVM intermittently. I moved the update to the rendering-thread instead. 

I'm not sure if the`setCurrentCommandNumber` is called from a different thread though. This would mean we could have some race condition?

ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/renderables/GcodeModel.java
```diff
@@ -314,32 +315,32 @@ private void updateVertexBuffers() {
...
-                    lineVertexData[vertIndex++] = (float)p2.z;
+                    lineVertexData[vertIndex++] = (float) p2.x;
+                    lineVertexData[vertIndex++] = (float) p2.y;
+                    lineVertexData[vertIndex++] = (float) p2.z;
                 }
             }
-            
+
             this.colorArrayDirty = true;
             this.vertexArrayDirty = true;
```
winder: Maybe adding these as another early exit condition would fix the race condition? These arrays may only be used when updating the render, which would mean we could be modifying them a second time while the first update is being applied.

ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/renderables/GcodeModel.java
```diff
@@ -158,13 +155,18 @@ public void draw(GLAutoDrawable drawable, boolean idle, Point3d machineCoord, Po
...
         // Batch mode if available 
+        boolean forceOldStyle = false;
         if(!forceOldStyle
                 && gl.isFunctionAvailable( "glGenBuffers" )
                 && gl.isFunctionAvailable( "glBindBuffer" )
                 && gl.isFunctionAvailable( "glBufferData" )
                 && gl.isFunctionAvailable( "glDeleteBuffers" ) ) {
             
             // Initialize OpenGL arrays if required.
+            if (this.vertexBufferDirty) {
```
winder: So this would be VertexBufferDirty && !vertexArrayDirty &&!colorArrayDirty

ugs-platform/ugs-platform-visualizer/src/main/java/com/willwinder/ugs/nbm/visualizer/renderables/GcodeModel.java
```diff
@@ -158,13 +155,18 @@ public void draw(GLAutoDrawable drawable, boolean idle, Point3d machineCoord, Po
         GL2 gl = drawable.getGL().getGL2();
```
winder: This might be a bigger problem, I wonder if the whole renderable pipeline needs to be fixed somehow. Whatever initiates the loop to call draw on all the renderable a should be able to skip frames like you did in updateVertexBuffers

winder:
@breiler it looks like you reverted the `invokeLater` portion. Does this still solve the event lagging?

breiler:
The invokeLater caused a bunch of problems with the JVM crashing originating from GL-bindings. When putting the updateVertexBuffer in the same thread as the rendering the event lagging problem disappeared. 

I couldn't see any problems with the 3D rendering. The max FPS is set to 15 which gives us ~60 ms  to do calculations. When testing Bobs big file the updateVertexBuffer took around 25 ms so I thought that it would leave plenty of time to calculate and render other stuff. And in a worst case scenario we would get a lower FPS.

PR #1378, opened 2020-04-01
Now stores if the visualiser window should open on startup

### LINKED ISSUE #1310: Visualizer and Welcome page "won't stay off"
## Problem description 
On windows 10, closing the windows for the visualizer and welcome page aren't "sticky" - when restarting the UGS they open again.

### Expected Behavior
Off means off. I should not need to close them every time I open UGS. Positions and open/close status of other pages seems to work properly in this respect.

### Actual Behavior
They open every time.

  1. Open the program
  2. Close the welcome page and the visualizer page
  2 . Close UGS
  3. Start UGS, the welcome page and visualizer page are still open. 
    
  UGS Platform 2.0 - Nightly Build Nov 2, 2019

#### Operating system
  Windows 10

#### Platform
  GRBL 1.1f

breiler:
I might have misunderstood how the setting for opening the Welcome window is supposed to function.

Removed the setting for "Show on startup" and instead use the built in function in netbeans which will remember if the user had the window open.

The same goes with the visualiser.

Fixes #1310


breiler:
Then I didn't misunderstand. =)

I remember having problems with the Welcome window starting even though I've unchecked the "Show on startup". But I couldn't reproduce it now so I'm guessing I was confusing it with the `mvn clean install` which clears my settings. I reverted the changes to the Welcome window.

PR #1385, opened 2020-04-08
Added joystick settings

breiler:
Added settings in the joystick plugin making it possible to change mappings of the controller buttons and axises. The actions that can be mapped comes from Netbeans. Added a live preview of the pressed buttons.

Screen Recording 2020-04-07 at 16 46 00 2020-04-07 16_51_07


jiffyjeff:
Thanks for the work on this!  I built this branch and tested using a Logitech Rumblepad 2.  Two issues stand out:

1. The Y and Z axis directions were reversed for the analog joysticks.  (Digital dpads were fine.)  I see an intentional flip of Y values being done in jamepad's ControllerIndex.getAxisState().  There's a related github issue: https://github.com/williamahartman/Jamepad/issues/14 
2. The size of the dead zone is not large enough, which leads to the jog not settling at rest. 

Again, thanks for your effort!  If you need any help, I'll be happy to troubleshoot/debug.

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/model/JoystickControl.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * The available buttons and axises that can be read by the service.
+ * The controls are divided into two groups:
+ *  - Analog - which will read analog values between 1.0 to -1.0
+ *  - Digital - which will read digital values such as buttons true/false
+ *
+ * @author Joacim Breiler
+ */
+public enum JoystickControl {
```
winder, line 33: This should support a lot of controllers! One thing I wonder is how it could support controllers with a digital D-Pad?

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/model/JoystickControl.java
```diff
@@ -0,0 +1,84 @@
...
+    A("platform.plugin.joystick.a", false),
+    B("platform.plugin.joystick.b", false),
+    X("platform.plugin.joystick.x", false),
+    Y("platform.plugin.joystick.y", false),
+    BACK("platform.plugin.joystick.back", false),
+    GUIDE("platform.plugin.joystick.guide", false),
```
winder, line 39: I've never heard of the guide button, is it the same as select? After some searching, it looks like this is an xbox thing?

It's probably safe to borrow the terminology used by retroarch. I think the changes are:
GUIDE -> SELECT
LEFT_BUMPER -> L
LEFT_TRIGGER -> L2
LEFT_STICK -> L3
RIGHT_BUMPER -> R
RIGHT_TRIGGER -> R2
RIGHT_STICK -> L3

https://retropie.org.uk/docs/RetroArch-Configuration/#example-default-per-system-retroarchcfg
https://github.com/libretro/RetroArch/blob/a96ae6e987671da71be4aa840e80eb341d4d62f3/retroarch.cfg#L491
```
# input_player1_a_btn =
# input_player1_b_btn =
# input_player1_y_btn =
# input_player1_x_btn =
# input_player1_start_btn =
# input_player1_select_btn =
# input_player1_l_btn =
# input_player1_r_btn =
# input_player1_left_btn =
# input_player1_right_btn =
# input_player1_up_btn =
# input_player1_down_btn =
# input_player1_l2_btn =
# input_player1_r2_btn =
# input_player1_l3_btn =
# input_player1_r3_btn =
```

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/service/JoystickServiceImpl.java
```diff
@@ -36,29 +44,88 @@ This file is part of Universal Gcode Sender (UGS).
...
+
+        actionManager = new ActionManager();
+        actionManager.registerAction("continuousJogXAction", "Actions/Machine", new AnalogJogAction(continuousJogWorker, Axis.X));
+        actionManager.registerAction("continuousJogYAction", "Actions/Machine", new AnalogJogAction(continuousJogWorker, Axis.Y));
+        actionManager.registerAction("continuousJogZAction", "Actions/Machine", new AnalogJogAction(continuousJogWorker, Axis.Z));
+
+        joystickActionDispatcher = new ActionDispatcher(actionManager, continuousJogWorker);
+        addListener(joystickActionDispatcher);
+
+        if (Settings.getVersion() == 0) {
```
winder, line 109: Could you add a comment here that says something like... "Check if the settings have not been initialized".

Probably not the most common edge case, but this does get a little tricky if multiple controllers are used or the controller is upgraded. It's probably fine to ignore this, since the user would be the only one who knows that the device has changed.

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/service/JoystickServiceImpl.java
```diff
@@ -133,14 +209,24 @@ private void readData() {
...
+            listener.onUpdate(joystickState);
+        });
     }
 
     private void updateJoystickAxisState(ControllerAxis controllerAxis) {
         try {
-            // Round values, got issues with the controller having 0.01 as zero-value
-            float value = Math.round(currentController.getAxisState(controllerAxis) * 10) / 10f;
-            JoystickAxis axis = Utils.getJoystickAxisFromControllerAxis(controllerAxis);
+            // We might have rounding errors from the controller, ignore the low value range
```
winder, line 223: Have you noticed that the analog inputs can change even if the joystick isn't moving? Would it help to keep a moving average?

winder: That would probably be part of the `currentController` for analog fields.

ugs-platform/ugs-platform-plugin-joystick/src/main/resources/resources/MessagesBundle.properties
```diff
@@ -0,0 +1,26 @@
+platform.plugin.joystick.activate = Activate joystick
```
winder, line 1: I'm still confused about how these work. Is the per-plugin message bundle overridden by the global message bundle?

ugs-platform/ugs-platform-ugslib/src/main/java/com/willwinder/ugs/nbp/lib/services/ActionRegistrationService.java
```diff
@@ -144,4 +155,87 @@ public void overrideActionName(String category, String key, String name) {
             Exceptions.printStackTrace(ex);
         }
     }
+
```
winder, line 158: Nice! The NetBeans people might like to hear about this: https://netbeans.apache.org/wiki/DevFaqActionsAddAtRuntime.asciidoc

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/Settings.java
```diff
@@ -30,12 +31,31 @@ This file is part of Universal Gcode Sender (UGS).
  */
 public class Settings {
     public static final String SETTINGS_ACTIVE = "active";
+    public static final String SETTINGS_VERSION = "version";
 
     private static Preferences preferences = NbPreferences.forModule(JoystickService.class);
```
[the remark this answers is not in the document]
breiler: Yes they are persisted. The downside is that the "Apply" button has little use as the settings are made instantly. But I figure that this can be added later when we have tried these features.

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/Utils.java
```diff
@@ -20,54 +20,66 @@ This file is part of Universal Gcode Sender (UGS).
...
+    public static final String ACTION_STOP = "Actions/Machine/com-willwinder-ugs-nbp-core-actions-StopAction.instance";
 
     /**
      * Translates from Jamepad controller button to a JoystickButton.
      *
      * @param controllerButton the controller button to translate from
      * @return a joystick button if it could be translated, or throws an illegalargumentexception
      */
-    public static JoystickButton getJoystickButtonFromControllerButton(ControllerButton controllerButton) {
+    public static JoystickControl getJoystickButtonFromControllerButton(ControllerButton controllerButton) {
```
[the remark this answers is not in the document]
breiler: I just wanted to separate our code from the Jamepad. This way we could can change the library without changing any of our implementation.

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/model/JoystickControl.java
breiler: Right now the support is limited. The buttons can't be mapped to a continuous jogging for instance. So if you want to jog it will only allow single stepping.

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/model/JoystickControl.java
breiler: Thanks, I'll update these!

ugs-platform/ugs-platform-plugin-joystick/src/main/java/com/willwinder/ugs/nbp/joystick/service/JoystickServiceImpl.java
breiler: I have the threshold at 3 percent which was required for my controller. Another user reported that this wasn't enough for their controller. This should probably be configurable as well or if we could come up with a clever algorithm for this as you suggested.

ugs-platform/ugs-platform-plugin-joystick/src/main/resources/resources/MessagesBundle.properties
breiler: No you are right. I used the MessagesBundle as the module bundle before, but this doesn't work anymore with the latest version of Netbeans. I removed this.

breiler:
@jiffyjeff thanks for trying this out! I made some updates:

> 1. The Y and Z axis directions were reversed for the analog joysticks. 

Added the possibility to reverse the axises.

> 2. The size of the dead zone is not large enough, which leads to the jog not settling at rest.

Added a setting for the zero threshold.

PR #1388, opened 2020-04-15
RunFromAction - Skip to specified line number

winder:
Backend components for "Run from..." line number action.

The preprocessor removes all lines until the specified line number is reached. For that line it rewrites the preamble to initialize the gcode state, and move the machine to the new start location.

Next I'll make some UI to visually select the line of code (either with the visualizer, or an editor action).


breiler:
Excellent work Will, the code is clean and simple!

PR #1430, opened 2020-07-29
Netbeans and JSerialComm upgrade

breiler:
Upgraded to the latest Netbeans release 12.1 which now supports hidpi scaling of icons. Because of this all SVG-icons needed to be reworked so that they had the right size and exists in all different variant 16, 24, 32, 16_dark, 24_dark, 32_dark, 16_dark_disabled, 24_dark_disabled, 32_dark_disabled (sigh)... The icon script fixes the rest of the variants as long as 16, 24, 32 icons are in place.

Needed to change the way the connection toolbar was created to support the built in dark look and feel.

Also upgrade jserialcomm library.


ugs-platform/ugs-platform-ugscore/src/main/java/com/willwinder/ugs/nbp/core/actions/PortAction.java
```diff
@@ -132,9 +114,34 @@ else if (evt.isStateChangeEvent()) {
             c.setVisible(!backend.isConnected());
         }
     }
-    
+
     @Override
     public Component getToolbarPresenter() {
+        if (c == null) {
```
winder, line 120: Does this change make startup faster? Should this function be synchronized in case it gets called twice?

breiler: There is only one toolbar allowed at a time so I think that this is safe.

PR #1445, opened 2020-10-06
Version string from version number

breiler:
Now displays the maven version in title:
Screen Shot 2020-10-06 at 21 00 23

If it's a release, only the version number will be displayed. If it's a snapshot build the the date will be shown as well.


CrazeUK:
Hi.  I am tryng to find someone to message re an issue I have found.

In the latest nightly build, no icons are appearing within the application.

Also, with the Run From command, it would be good to show you what line the previous line was stopped on.
Oh also, when you hit pause, can the Z axis be lifted and the spindle stopped.?

winder (approved):
Looks good to me, sorry for the delay!

Issue #1249, opened 2019-07-04
Confirmation of "Reset Zero"


AngeloQABR:
Hi!
"Reset Zero" button is just near "Return Zero" button... I know, is my problem but I clicked "Reset Zero" button for two times when I wanted return to zero.
Is it possible create a message for confirmation in "Reset Zero" button? It would have saved me on the two jobs I lost.
Thanks!

AbySet:
This happend to me too, but since it is not a bug i did not post it.
I should say i totaly agree with you @AngeloQABR , any improuvement will be welcome or at least colored text in red or orange but confirmation is better.
Thanks to you and to everyone working on this project.

jahnj0584:
Quickest solution is to make these commands as a seperate macro, which you can easily re-order and seperate from each other

TekMason:
+1 for this feature built-in.

breiler:
It is now possible to reorder the buttons in the toolbox window:
https://github.com/winder/Universal-G-Code-Sender/wiki/Usage#toolbox

Issue #933, opened 2018-03-05
Add button to activate sleep mode


Harvie:
Grbl 1.1 has feature, that it can go to sleep mode using $SLP command and wake up by reset.
Can you please add button for this? There's place just for one more button :-)

![image](https://user-images.githubusercontent.com/276504/36956964-7cf5f818-2031-11e8-90a3-3b3033681bb6.png)

jahnj0584:
Can you add it as a macro?

yaroslavVl:
Could you please add SLEEP button?!

breiler:
This can be added as a macro

Issue #219, opened 2015-01-30
Vac and Coolant Buttons


shmorgan:
Add a Vac and Coolant On/Off button on the Machine Control Page.

philreindl:
These could be set up as macros.  There are now macro buttons on the machine control page, so no tabbing back and forth would be needed to use them.

breiler:
This can be done in the platform and classic edition using macros.

PR #1458, opened 2020-10-18
Added a new toolbox component with customizable actions

breiler:
This is new plugin for displaying actions in UGS. It has a reactive design and allows users to customize which actions that should be available.

Adresses these issues:
#219, #933, #1249

Default view when component is loaded the first time:

Is resizable and will adjust the number of columns:

---

It also allows customizing the actions from a settings panel:
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,6 @@
 
     <!-- cloud storage plugin -->
     <ugs.s3-filesystem.version>2.2.2</ugs.s3-filesystem.version>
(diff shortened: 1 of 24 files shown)


winder (approved):
This looks great. My only comment is that we aren't really using Jackson before this. I don't have a strong attachment to Gson, but since it's used elsewhere I would probably use it here.

breiler:
@winder thanks, Gson is a lot cleaner to work with. Thanks!

PR #1460, opened 2020-10-22
Bugfix/reset zero bug

### LINKED ISSUE #1420: "Reset Zero" button not working properly in inch mode
## Problem description 
"Reset Zero" button not working properly in inch mode.

### Expected Behavior
Axes are zeroed

### Actual Behavior
Inch position is converted to mm and then fed to controller without switching controller to mm mode.

Example:
Machine is in inch mode (G20)
Position is x=0, y=0, z=-0.5
DRO shows x=0, y=0, z=-0.5
"Reset Zero" button is pushed
`G10 L2 P1 X0 Y0 Z-12.7` sent to controller
DRO shows x=0, y=0, z=12.2
Machine is still in inch mode

  1. Open the program
  2. Connect to control board
  3. Make sure you are in G20 (inch) mode
  4. Z down 0.5"
  5. Click "Reset Zero" button in "Common Actions" window
  

#### Version
UGS Platform 2.0 - Nightly Build Jun 25, 2020
 

#### Operating system
Linux on RaspberryPI

#### Platform
g2core

breiler:
Fixed problem with positions being used with unknown unit. Removed constructor and methods for creating PartialPosition without units to prevent these kinds of bugs reappearing.

Fixes #1420

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/AbstractController.java
@@ -164,7 +164,7 @@ public abstract class AbstractController implements CommunicatorListener, IContr
     public void returnToHome(double safetyHeightInMm) throws Exception {
         if (isIdle()) {
             // Convert the safety height to the same units as the current gcode state
-            UnitUtils.Units currentUnit = UnitUtils.Units.getUnits(getCurrentGcodeState().units);
+            UnitUtils.Units currentUnit = getCurrentGcodeState().getUnits();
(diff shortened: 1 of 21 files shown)


winder (approved):
Good idea, this should fix the problem and prevent accidents in the future.

My only suggestions are design related (and optional):
* `PartialPosition.Builder()` could be a static function to let you skip the `new`
* Maybe `Build` could be `Build(Units)` to make it even harder to forget to set units.
* Alternatively `build()` could throw an exception if the units are still default.

PR #1466, opened 2020-11-05
Added a editor cursor that is positioned on the currently selected line

breiler:
Added a editor cursor that is positioned on the currently highlighted line. This will be helpful when selecting "Run from..." in the editor.


winder (approved):
Looks great

Issue #1504, opened 2021-01-01
DRO usability issues


breiler:
Today when UGS platform is opened for the first time the windows needs to be moved to able to see them properly. I would like to rearrange some windows to make the user experience for new users a bit better and also make room for new functions:
![image](https://user-images.githubusercontent.com/8962024/103439575-66b8c300-4c3e-11eb-85ae-07159e0e6d15.png)

I have a couple of proposal that I would like to get your input on:
1. Make the DRO able to scale down event more. This is needed to support really small screens such as the Raspberry Pi 7" touch screen. The LED-font needs to be replaced as it doesn't scale down well.
![image](https://user-images.githubusercontent.com/8962024/103439676-59e89f00-4c3f-11eb-92d3-534d1bc72a0e.png)

2. Add a new popup editor that will work a bit better then todays inplace editor which shows the wrong colors on Linux and Windows:
![image](https://user-images.githubusercontent.com/8962024/103439716-97e5c300-4c3f-11eb-9e52-014ddd13c008.png)

3. Add three new icons to the toolbar so that the toolbox window isn't needed to be shown the whole time. I figure we need at least three icons for Homing, Unlocking and Restarting the controller:
![image](https://user-images.githubusercontent.com/8962024/103439741-cfed0600-4c3f-11eb-8991-1b98dad01000.png)

4. Remove one of the leftmost rows so there are typically just two rows:
![image](https://user-images.githubusercontent.com/8962024/103439809-7507de80-4c40-11eb-88fc-c81b0107db35.png)

PR #1508, opened 2021-01-07
Adjustments to the DRO

breiler:
This PR contains small adjustments to make room for further improvements and features.

1. Made the DRO able to scale down more. This is needed to support really small screens such as the Raspberry Pi 7" touch screen. The LED-font needs to be replaced as it doesn't scale down well.
image

2. Added a new popup editor that will work a bit better then todays inplace editor which shows the wrong colors on Linux and Windows:
image

3. Added three new icons to the toolbar so that the toolbox window isn't needed to be shown the whole time. I figure we need at least three icons for Homing, Unlocking and Restarting the controller:
image

4. Remove one of the leftmost rows so there are typically just two rows:
image

Fixes: #1504


winder (approved):
Looks great!

PR #1515, opened 2021-01-16
Add ABC Actions and jog settings.

winder:
Next set of building blocks for ABC support. This was all mostly busy work following the existing patterns and adding the new axes:
- actions - in the future these should be enabled/disabled based on controller capabilities.
- ABC step size settings
- updates to jog panel for configuring ABC step size
- minor improvement to the jog panel so that enable/disable Z/ABC step sizes don't leave empty cells in the layout.

image

These are also in the menu
image
image


ugs-core/src/com/willwinder/universalgcodesender/utils/Settings.java
```diff
@@ -256,6 +267,15 @@ public void setzJogStepSize(double zJogStepSize) {
         changed();
     }
 
+    public double getabcJogStepSize() {
```
breiler, line 270: Really minor, i would prefer `getABCJogStepSize()`

breiler (approved):
Looks good!

PR #1519, opened 2021-01-20
Open Gcode editor by default

breiler:
This PR contains changes so that the editor is opened when a file is loaded. It will only allow one editor at a time.

My intention is to add a designer tool with its own mime type which will be shown instead of the editor window and to use netbeans DataObjects for this.

I also found some issues with the jog component.


winder (approved):
Looks good

Issue #1507, opened 2021-01-07
Visualizer not updating when making changes to Gcode, file not being send when it is updated


jimmys01:
**Describe the bug**
When you change/add/remove gcode commands from an opened file the visualizer does not update accordingly.
Also the file is not send to the controller. You have to save the file and the re-open it to mitigate this.

**To reproduce**
Open a gcode file
On the editor add commands
the commands are not send to the controller and also are not vizualized

**Expected behavior**
Update the visualizer real time or near real time 
and send the file to the controller each time it has changed and you press the green Send button

**Version**
UGS Platform 2.0.7 windows 10

**Hardware**
GRBL 1.1f ESP32

Windows 10

breiler:
I never noticed this, wondering if this could be a Windows issue:
![2021-01-07 19-03-10 2021-01-07 19_04_35](https://user-images.githubusercontent.com/8962024/103927864-48aa0180-511b-11eb-8dc8-892149fd1cd9.gif)

Could you post the log file, it might have some clues:
https://github.com/winder/Universal-G-Code-Sender/wiki/Configuration#log-files

jimmys01:
You are amazing fast :-). In the video you posted the behavior is right. 
[ugs.zip](https://github.com/winder/Universal-G-Code-Sender/files/5783242/ugs.zip)
I am fooling arount a lot with UGS I hope there are not many trash in the logs...

breiler:
The logs were filled with errors but I can't find any trace of our code being faulty. This looks suspicious though:
```
	at java.desktop/javax.swing.RepaintManager.paint(Unknown Source)
	at java.desktop/javax.swing.JComponent._paintImmediately(Unknown Source)
	at java.desktop/javax.swing.JComponent.paintImmediately(Unknown Source)
	at java.desktop/javax.swing.RepaintManager$4.run(Unknown Source)
INFO [null]: Last record repeated again.
	at java.base/java.security.AccessController.doPrivileged(Unknown Source)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
	at java.desktop/javax.swing.RepaintManager.paintDirtyRegions(Unknown Source)
INFO [null]: Last record repeated again.
```

I am wondering if this could be caused by a antivirus program or you could try running UGS with administrator rights?

jimmys01:
[...]

PR #1520, opened 2021-01-21
Fixed problem with opening files on windows

breiler:
The GcodeDataObject is not recreated on windows if opening the same file twice. Needed to move the opening code to the SourceMultiViewElement instead and listen to if the TopComponent is opened.

Should fix the issue: #1507


jimmys01:
It will update the visualization when I go to File->save but not in real time.
For my work case this is fine but it is not the same as you show in your first video

PR #1534, opened 2021-02-08
Bugfix/visualize rotations

breiler:
I think I solved the problem with rotational model from Bob and the train shaft model from #1068

You were so close Will 😃 

I should probably write some unit tests but its getting late.


winder (approved):

winder:
Amazing!
image

The driveshaft model looks a little strange but ncviewer.com renders it the same way so maybe that is by design
image

PR #1540, opened 2021-02-15
First pass at 6-Axis GRBL ESP32 support.

winder:
Added some minimal support for GRBL ESP32, along with DRO integration.

* New firmware in the dropdown.
* Add ABC to DRO
* Controllers are now all setting XYZ capabilities. GRBL ESP32 adds ABC if they are enabled.
* DRO now uses capabilities to enable the different axes.
* Updated the existing grbl status parser to look for optional ABC coordinates.

Issues I've noticed so far:
* sometimes the status message seems to be missing the leading `<` and it ends up being printed to the console. I'm not sure if thats a UGS, Grbl_Esp32, or my ESP32 board.
* Running the TrainDriveShaft file shared in another issue reports A values much larger than I see in the file. This could be an issue with Grbl_Esp32, or maybe UGS is doing some processing on A values that it shouldn't.

image

image


breiler (approved):
I think that this looks good and I could also give it a try later tonight to see if I have the same issues as you have with the status reports.

The only concern I have is the units for the rotational axis. Does it report in degrees which would be converted between metric and imperial units. But I think that this should work for now.

breiler:
My concern regarding the ABC units was already taken care of. I ran some of the other models which seemed to work as intended. I did not get any incomplete responses from the controller and the models I tried with ran fine.

This PR looks good to me

rlwoodjr:
Would it be possible to select the DRO Axes we want to display?  It would be nice not to display axes that are not needed.

winder:
> Would it be possible to select the DRO Axes we want to display? It would be nice not to display axes that are not needed.

Yes, thats the plan. My next step is to hide unsupported axes, after that would be adding a way to hide any supported-but-not-used axes.

winder:
It took a little longer than I expected, but everything seems to be working pretty nicely. The disabled axes are now hidden and you can override which ones are visible with a right-click menu or in the settings. I'm not sure how useful the right-click menu, maybe it should be removed now that development is finished..

Right click menu + XZA configuration:
image

Options:
image

breiler (approved):
Nice work!

PR #1547, opened 2021-02-26
flakiness in bucket

ChopinLi-cp:
The test com.willwinder.ugs.nbp.S3FileSystemViewTest.testGetFiles asserts that under the directory "s3:/test-files/" there
are two files. However, the createBucket method has a caching method that saves the bucket with the same name across tests. So if another test like
com.willwinder.ugs.nbp.S3FileSystemViewTest.testUploadDownloadFile calls the method createBucket with the same name
and then puts files under this bucket, it will pollute the cached bucket and the assertion will fail.


winder:
Thanks for the contribution!

PR #1557, opened 2021-03-14
Marlin support

breiler:
Let us have a look if we can add this. 

I had a bunch of problems when I tried building support for marlin in my [branch](https://github.com/winder/Universal-G-Code-Sender/compare/master...breiler:feature/marlin-firmware-2), and there seems to be a couple of small problems with  @andy-hnk's version as well (i.e capabilities are missing).


ugs-core/src/com/willwinder/universalgcodesender/MarlinCommunicator.java
```diff
@@ -0,0 +1,90 @@
...
+		}
+		return super.allowMoreCommands();
+	}
+
+	protected MarlinCommunicator() {
+	}
+
+	@Override
+	public int getBufferSize() {
+		return 200; // TODO: what should it be?
```
breiler, line 50: Browsed through the marlin code and it looks like they use 128 as default:
https://github.com/MarlinFirmware/Marlin/blob/8e1ea6a2fa1b90a58b4257eec9fbc2923adda680/Marlin/src/inc/Conditionals_adv.h

In GRBL we used 123 as we needed some space for real time commands and line breaks.

ugs-core/src/com/willwinder/universalgcodesender/MarlinController.java
```diff
@@ -0,0 +1,409 @@
...
+		GcodeCommand command = createCommand(commandString);
+		command.setTemporaryParserModalChange(true);
+		sendCommandImmediately(command);
+		// restoreParserModalState();
+	}
+
+	/**
+	 * Begin issuing GRBL status request commands.
+	 */
+	private void beginPollingPosition() {
```
breiler, line 268: Much of this code can now be replaced with the [StatusPollTimer](ugs-core/src/com/willwinder/universalgcodesender/StatusPollTimer.java).

ugs-core/src/com/willwinder/universalgcodesender/MarlinController.java
```diff
@@ -0,0 +1,409 @@
...
+
+				// this.handleStatusString(response);
+
+				controllerStatus = MarlinUtils.getStatusFromStatusString(
+						controllerStatus, response, capabilities, getFirmwareSettings().getReportingUnits());
+
+				dispatchStatusString(controllerStatus);
+
+				this.checkStreamFinished();
+			} else if (response.contains("Free Memory:")) {
```
breiler, line 198: Is this the best way to identify that it is connected to a Marlin controller?

When I experimented with this I used the "start" message:
https://github.com/winder/Universal-G-Code-Sender/compare/master...breiler:feature/marlin-firmware-2#diff-4efdcf4b045d1fc9c9610b760d4a7589fa3a7ab94273ea524d6b8821e64c4b65R109

Issue #1564, opened 2021-03-22
Arduino Micro not recognized as Joypad


mvezzelli:
**Describe the bug**
I created a custom hardware pendant connected to an Arduino Micro in Joypad mode. On Windows, joypad device is listed and working as expected. On UGS nothing happens if I move axis.

**To reproduce**
Connect Arduino Micro in joypad mode, move axis press buttons

**Expected behavior**
Mapped joystick event is triggered

**Version**
2.0.7 build 20210320, same on latest release

**Hardware**
GRBL 1.1f

**Operating system (please complete the following information):**
Windows 10,

**Additional context**
The controller is a standard 5 axis, 32 buttons, 1 POV controller.
I do believe it's very common that people are trying to use Arduino Micro/Leonardo to create their own pendants.
![immagine](https://user-images.githubusercontent.com/23317232/111983860-e56d2000-8b0a-11eb-9663-0f4ba4815d4f.png)

breiler:
We are using the GameControllerDB for identifying different game controllers. We can add new custom controllers if we have an identification string for it. Use one of GameControllerDB:s tools to extract the unique string for your controller and we can add it to the configuration file: https://github.com/gabomdq/SDL_GameControllerDB

mvezzelli:
Thanks for the quick reply! I generated this, is there a way to test it before adding?

03000000412300003780000000000000,Arduino Micro,platform:Windows,a:b0,b:b1,x:b2,y:b3,back:b4,guide:b5,start:b6,leftstick:b7,rightstick:b8,leftshoulder:b9,rightshoulder:b10,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,-leftx:a0,+leftx:a1,-lefty:a2,+lefty:a3,-rightx:a4,+rightx:a5,-righty:a6,-righty:a7,lefttrigger:b11,righttrigger:b12,

breiler:
Awesome, I've added a new configuration and it is building, you can check its status here: https://travis-ci.org/github/winder/Universal-G-Code-Sender/builds/763967416

Once that is finished you can download it from the latest nightly build here: https://github.com/winder/Universal-G-Code-Sender#downloads

breiler:
Let us know if this worked. Also, I would love to see a demo! =)

mvezzelli:
Thanks for the quick fix, I need some days to finalize then I'll be happy to post a demo :)

claudio-wilson:
[...]

PR #1565, opened 2021-03-22
Added custom arduino joystick configuration

breiler:
Adresses issue #1564


hmantz:
How can I add different custom joystick configurations? Do I have to edit gamecontrollerdb.txt and compile the entire code again every time I add a new one?  Is it possible to have gamecontrollerdb.txt (or some other text configuration file) in the settings directory and be able to edit it with a text editor?

Thanks

Issue #1584, opened 2021-04-20
Build dependency problems with ugs-core.jar on latest UGS source code on master


lamikr:
I try to build the latest 2.0.7-SNAPSHOT from master branch with Maven on console and in the end of the build I am seeing errors for loading some classes that are included in ugs-core.jar

INFO] ugs-platform-plugin-toolbox ........................ SUCCESS [  1.803 s]
[INFO] ugs-platform-plugin-dro ............................ SUCCESS [  3.084 s]
[INFO] ugs-platform-app ................................... FAILURE [ 11.862 s]

And it's caused by following type of error.
ERROR]   Run 2: ApplicationTest.testApplication NbModuleSuite has been started with failOnMessage(WARNING) and failOnException(INFO). The following failures have been captured:
[org.netbeans.ProxyClassLoader] THREAD: main MSG: Will not load class com.willwinder.universalgcodesender.model.UnitUtils$Units arbitrarily from one of ModuleCL@0[com.willwinder.ugs.platform.ugslib] and ModuleCL@1[com.willwinder.ugs.platform.plugin.joystick] starting from SystemClassLoader[133 modules]; see http://wiki.netbeans.org/DevFaqModuleCCE
[org.openide.filesystems] THREAD: main MSG: null
java.io.IOException
  org.openide.filesystems.XMLMapAttr$Attr.decodeValue(XMLMapAttr.java:865)
  org.openide.filesystems.XMLMapAttr$Attr.getObject(XMLMapAttr.java:944)
  org.openide.filesystems.XMLMapAttr$Attr.get(XMLMapAttr.java:661)
  org.openide.filesystems.XMLMapAttr.getAttribute(XMLMapAttr.java:226)
  org.openide.filesystems.XMLMapAttr.getAttribute(XMLMapAttr.java:191)

1. git clone https://github.com/winder/Universal-G-Code-Sender.git
2. mvn -X install
3. And then the error happens in the end of the build
5. **Expected behavior**
Build would work without error.

**Version**
I am building on Linux and my java-version is
openjdk version "11.0.10-ea" 2021-01-19 LTS
OpenJDK Runtime Environment 18.9 (build 11.0.10-ea+1-LTS)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.10-ea+1-LTS, mixed mode, sharing)

**Additional context**
I am however able to launch the ugs-platform but then I am seeing a similar type of clasload errors happening to some other
[...]

PR #1586, opened 2021-04-26
Fix ClassNotFoundException classpath error

lamikr:
Fix the problem for running tests during the build of ugs-platform-app.

[INFO] ugs-platform-plugin-dro ............................ SUCCESS [  4.410 s]
[INFO] ugs-platform-app ................................... FAILURE [ 17.721 s]

Failure is caused by the error that occurs because two different modules
that depends from each other have same ugs-core jar in the classpath
and this causes error because the class is loaded from the wrong
classpath. Exact error printed out by the netbeans classloader:

Caused by: java.lang.ClassNotFoundException: Will not load class
com.willwinder.universalgcodesender.model.UnitUtils$Units arbitrarily
from one of
ModuleCL@0[com.willwinder.ugs.platform.ugslib] and
ModuleCL@1[com.willwinder.ugs.platform.plugin.joystick]
starting from SystemClassLoader[133 modules];
see http://wiki.netbeans.org/DevFaqModuleCCE
  org.netbeans.ProxyClassLoader.doFindClass(ProxyClassLoader.java:191)
  org.netbeans.ProxyClassLoader.loadClass(ProxyClassLoader.java:125)

Fixes https://github.com/winder/Universal-G-Code-Sender/issues/1584

Signed-off-by: Mika Laitio <lamikr@gmail.com>

### DIFF
--- a/ugs-platform/ugs-platform-plugin-dro/pom.xml
+++ b/ugs-platform/ugs-platform-plugin-dro/pom.xml
@@ -37,6 +37,7 @@
             <artifactId>ugs-core</artifactId>
             <version>${project.version}</version>
             <type>jar</type>
+            <scope>provided</scope>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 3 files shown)


lamikr:
This allows the "mvn install" command to pass without errors. (no need to use -DskipTests=true parameter)
See error issue: https://github.com/winder/Universal-G-Code-Sender/issues/1584

breiler:
Looks good, thanks!

Issue #1409, opened 2020-06-16
Add simple a simple vector graphics editor


breiler:
## Feature request
Universal G-code Sender has the primary focus to send existing gcode to the hardware. The CAD/CAM has to be done using other software. For beginners this may seem daunting to have to learn so many tools and techniques.

By adding support for doing simple tasks in Universal G-code Sender such as generating tool paths for cutting or engraving using images, vector graphics or to just engrave text we could lower the threshold for beginners to start making things.

This is a feature that I am currently working on and is keeping me up at nights. I'll try to log my work progress in this issue and hopefully get some feedback and insights from you guys.

Here is a small screen recording of the progress right now (still a lot to do). Open an SVG file and configure the cut operation on the geometric shape.
![Screen Recording 2020-06-16 at 12 37 54 2020-06-16 12_44_09](https://user-images.githubusercontent.com/8962024/84765873-278f9100-afd0-11ea-9b55-42654b0aa826.gif)

Please leave a comment of what you think?

khturnings:
This is looking awesome. 
Sometimes it is such a pain having to go to another program or in my case another computer when you just want to do something very simple like cut a circle out or a quick text engraving or just run a test on your machine.

Happy to test this for you if you need some testing.
Great work.

timon2jkl:
I like this very much I also like for simple things and testing very nice for the newbies to get them started and pointed in the right direction

krosspaper:
This is a fantastic feature, can't wait for it!

breiler:
I just want to give an update on the progress on this. There is still a lot of issues left to handle before it is ready for release, but I think it is coming along and I'm pleased with how it is turning out. It maybe doesn't look that different from the PoC screen cast posted earlier, but it is almost completely rewritten and I'm doing my best to making a solid base to continue working on.

In this demo screen cast I'm showing the basic functionality implemented so far and how it is currently integrated in UGS with NetBeans API:s:
* Creating a new design document
* Adding primitive geometries, performing basic transformations and setting the cut operation to pockets (moving and rotating)
* Importing an SVG 
* Changing the tool settings (tool diameter and "step over")
* Saving and Open

![Ugs designer demo 2021-07-08 07_11_03](https://user-images.githubusercontent.com/8962024/124865835-b3999f80-dfbb-11eb-801f-96aeca63fdde.gif)

[...]

PR #1629, opened 2021-07-12
Feature/designer

breiler:
Sorry for the massive PR, but this is something that I've been working on for a long time now and I would appreciate if you could take a look at the broad picture.

I'm currently using the JTS-library for making geometric transformations, but I think that I can add that specific code for those operations later on and remove that library.

There is also some duplication with coordinates and axises which can polished to use the central ones in UGS.

Relates to #1409


winder:
This is amazing. I'll do my best to review, but maybe we should merge it in before that so that people can try it out right away?

khturnings:
> This is amazing. I'll do my best to review, but maybe we should merge it in before that so that people can try it out right away?

In line with this could this be packaged as a beta or development release so non-programmers can do some testing . I would love to give this a go.

breiler:
It should not cause any problems with the existing functionality by merging it now. 

But as I mentioned earlier this barely contains all the MVP features and there is a lot of work left to be done. Just so that you have the right expectations when trying this out. But I would really appreciate the help with testing and finding the most annoying bugs and usability pains.

PR #1635, opened 2021-07-20
First functionality implementation of PatternReplacer to improve the …

AndyCXL:
…original PatternRemover. Uses sed "s/regex/replacewith" for Replacer OR regex "regex" to give the original behaviour. No changes to the JSON storage of settings.


AndyCXL:
This Pull Request appears only to have incorporated one changed file, whereas there is one new file and one other changed that I cannot see. Not sure this has been successful...

AndyCXL:
Cyclic import reference between CommandProcessor and BackendAPI instance means retrieving Macros is not yet possible (all the rest of the code is done ready for the solution to this). Will look to consolidate the code so it performs the simple replacer function but is ready for the macro expansion the moment this cyclic issue can be worked around - then will re-submit the pull request

PR #1640, opened 2021-07-23
Implements #1158 Replacer to enhance Remover Processor

AndyCXL:
The macro expansion code is working and robust, using sed " s/regex/%macroname% " syntax and has remained a simple piece of code too. I have run into an import find issue which means I will make one small adjustment to the code to switch off the macro expansion but leave its code in place so that the Replacer function can be merged in the meantime if seen as progress.


AndyCXL:
Cyclic dependency remains an obstacle.

PR #1642, opened 2021-07-24
#1158 Enhance ProcessorRemover to Replacer

AndyCXL:
#1158 Enhance ProcessorRemover to Replacer. Macro expansion disabled (code commented out) until cyclic dependency issue can be overcome - will need someone with more architectural understanding to help address this.


AndyCXL:
It fails build here, but builds correctly without cyclic dependency (which is removed from the code) using the build script on my local machine. I will admit now to not knowing what is happening - over to someone with higher knowledge to point me/correct me/whatever.

AndyCXL:
Will take a fresh fork and try submitting again. Frustrating, but it is something to try...

PR #1644, opened 2021-07-26
Addresses #1158 Replacer functionality

AndyCXL:
Addresses #1158 Replacer functionality. Code for %macro% functionality is written too, just waiting for solution to backend instance before enabling. This Pull already delivers useful enhancement to the Remover function.


AndyCXL:
Want to see a more modest pull succeed before pursuing this. See #1645

Issue #1645, opened 2021-07-26
Menu Multiply XY Step Size by 10 not working


AndyCXL:
**Describe the bug**
Select Menu Multiply XY Step Size by 10, performs no action.
Not noticed any other such menu function omissions.

**Expected behaviour**
Expecting XY Step Size to x10. Divide by does work.

**Version**
UGS Platform 2.0.8-SNAPSHOT of 26-Jul-2021, 16:00 GMT+1

**Hardware**
GRBL 1.1t Mega 5X

**Operating system (please complete the following information):**
MacOSX

**Additional context**

testanley3:
This bug also present in Windows 10 edition of the same version.  Glad it's not just my imagination.

AndyCXL:
I have found the typo in the source that is causing this, and shortly will submit a pull request with the very simple fix.

testanley3:
Thank you for this fix.  Forgive my ignorance, but how do I update my copy of the program to implement this change?

AndyCXL:
Several ways;
1. Best is that my Pull Request is merged and this becomes part of the nightly builds
2. You fork my copy (Menu_Multiply from GitHub AndyCXL) compile and use that until 1. happens

AndyCXL:
@testanley3 if you can PM me, I can send the ZIP file I have compiled of either Menu_Multiply or Replacer branches. It's not ideal as you are then wholly dependent on me, but if you can't compile from my repository it may a practical 'fix'.

AndyCXL:
@testanley3 Note that this proposed fix (Menu Multiply XY Step Size by 10) has been accepted and merged into the nightly builds, so you can download a 'fixed' version of UGS

testanley3:
Thanks to everyone!

testanley3:
Hello. I have tried to load a nightly build that incorporates this Multiply
XY by 10 patch, but the latest I can find is dated 07/24/21 and this does
not seem to contain the correction. What am I doing wrong?

testanley3:
[...]

PR #1652, opened 2021-08-05
Replacer #1158 and Second Typo #1645

AndyCXL:
Replacer #1158 (PatternRemover enhancement) and Second Typo #1645 (ABC Step Multiply)

PatternRemover has been enhanced to support 'replace with' and %macro% expansion by introducing 'sed' syntax to the preferences/UGS/Contoller Options. 'sed' syntax is "s/regex to match/replacement text" so the following now all work;
- M6T1 in code, "M6T[0-9]+" enabled will REMOVE M6T1 as per original functionality
- M6T1 in code, "s/M6T[0-9]+/G21" enabled will REPLACE M6T1 with G21 if that is what the user desires
- M6T1 in code, "s/M6T[0-9]+/%1%" enabled will REPLACE M6T1 with whatever macro "1" is defined as by the user in the Gcode column. User must be aware that the result from macro 1 will be passed through as-is.
- All REMOVES or REPLACES are logged under INFO to assist user debugging regex and %macro% settings.
Confirmed working (Apple and RPi).

Found a typo in jog service.java where getStepSizeZ() was errantly against multiplyABCStepSize(), should have been getStepSizeABC(), which is corrected.
Confirmed working (Apple and RPi).


AndyCXL:
Not sure what, if anything, I am supposed to do with the CodeBeat response. Nothing wrong with the code aside from style and brevity as-per it's assessment???

AndyCXL (commented):
Tidied some comments to improve code style

AndyCXL:
As CodeBeat didn't like my coding style, I re-wrote the main PatternRemover loop to be more elegant - it is more legible and has fewer nesting levels. I also tested the Macro expansion against macros named '1' or 'Macro #1' as these could both exist, and it successfully identifies and expands either form.
Hopefully this can merge into the base now...

PR #1676, opened 2021-08-24
PatternRemover Enhancement #1158 and Second Typo #1645

AndyCXL:
Replacer #1158 (PatternRemover enhancement) and Second Typo #1645 (ABC Step Multiply)

PatternRemover has been enhanced to support 'replace with' and %macro% expansion by introducing 'sed' syntax to the preferences/UGS/Contoller Options. 'sed' syntax is "s/regex to match/replacement text" so the following now all work;

M6T1 in code, "M6T[0-9]+" enabled will REMOVE M6T1 as per original functionality
M6T1 in code, "s/M6T[0-9]+/G21" enabled will REPLACE M6T1 with G21 if that is what the user desires
M6T1 in code, "s/M6T[0-9]+/%1%" enabled will REPLACE M6T1 with whatever macro "1" is defined as by the user in the Gcode column. User must be aware that the result from macro 1 will be passed through as-is.
All REMOVES or REPLACES are logged under INFO to assist user debugging regex and %macro% settings.

Confirmed working (Apple and RPi). Core algorithm re-written to reduce if nesting depths.

Found a typo in jog service.java where getStepSizeZ() was errantly against multiplyABCStepSize(), should have been getStepSizeABC(), which is corrected.

Confirmed working (Apple and RPi).


AndyCXL:
Over to the moderators to accept, critique or reject this so I know what standard is expected.

AndyCXL:
This pull has sat untouched for 4 months, having been asked to make a minor change which was done and re-submitted for review. Are contributions like this really wanted, or not really?

breiler:
Sorry, it is of course nice with contributions. I do however not feel the ownership of what should be merged and what shouldn't, I feel that it is up to Will. I should have been more clear on where I stood.

This is a nice feature, my only concern is how well this plays together with other features. For instance when using macros there is a feature that can popup asking the user for inputs, what will happen if running this in a headless mode.

I know that this feature is more for expert users, so they would probably know what they are doing. So in that sense this PR looks good to me. I'll leave the merging for Will.

breiler (approved):

AndyCXL:
I can't tell if the Branch Conflict is actually outstanding or not, or what/if I am required to do to fix it.

AndyCXL:
I give up. Changing to a different sender app now.

PR #1678, opened 2021-08-26
This fixes a serial buffer overrun that can happen in some cases when…

bertieconfundo:
… a command has whitespace on it (ie, from a multi-command macro).  This also fixes the Grbl buffer size to be 128 - it was conservatively set at 123 with the assumption that space needed to be allocated for real-time commands. This is not true. Real-time commands do not consume ring buffer space. Updated the GrblCommunicator test to reflect the buffer size change.

Without this fix, I can reliably produce a buffer overrun by running a simple macro (G20; G53 G0 Z0;G53 G0 X0 Y2.6) 16 times and then running a previously-loaded gcode program.  The reason the macro has to be run a number of times is that the sent byte count gets increasingly off each time the macro is run (because of the .trim()'d command being used for the byte count when sending, but not using .trim() on the command when adjusting the byte count when the response is received.

With the fix in place, I cannot reproduce the buffer overrun.

The overrun was difficult to debug because it manifests as dropped characters - when the Grbl firmware receives a serial byte and the buffer is full, it's silently ignored (!!), which can end up causing a whole bunch of unpredictable behavior - weird error messages about badly formatted gcode, and even machine movements that were not intended resulting in crashes.  In order to help debug this, I modified the Grbl firmware slightly so that if there was ever a byte received when the ring buffer was full, it would mc_reset() and go into the ALARM state - this protects against executing corrupt code, and it made it a lot easier to diagnose and fix this problem.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/BufferedCommunicator.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/BufferedCommunicator.java
@@ -213,7 +213,7 @@ public abstract class BufferedCommunicator extends AbstractCommunicator {
                 continue;
             }
 
-            String commandString = command.getCommandString().trim();
+            String commandString = command.getCommandString();
             
             this.activeCommandList.add(command);
             this.sentBufferSize += (commandString.length() + 1);
--- a/ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/GrblUtils.java
@@ -39,8 +39,8 @@ import java.util.regex.Pattern;
  */
 public class GrblUtils {
     
-    // Note: 5 characters of this buffer reserved for real time commands.
-    public static final int GRBL_RX_BUFFER_SIZE= 123;
+    // Note: The Grbl RX buffer is not consumed by real-time commands
+    public static final int GRBL_RX_BUFFER_SIZE= 128;
 
     /**
      * Grbl commands
--- a/ugs-core/test/com/willwinder/universalgcodesender/GrblCommunicatorTest.java
+++ b/ugs-core/test/com/willwinder/universalgcodesender/GrblCommunicatorTest.java
@@ -212,7 +212,7 @@ public class GrblCommunicatorTest {
         MockConnection mc = new MockConnection(mg.in, mg.out);
         GrblCommunicator instance = new GrblCommunicator(cb, asl, mc);
         String term = "\n";
-        String thirtyNineCharString = "thirty-nine character command here.....";
+        String thirtyCharString = "thirty character command......";
 
         boolean result;
         boolean expResult;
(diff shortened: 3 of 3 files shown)


dvsingletary:
Yes. This is what I used when making the corrections in the related commit that was merged on 11/18/2021.

bertieconfundo:
Yes. I referenced the defined streaming protocol when making this fix.

J-Dunn:
OK, thanks. Nice work fixing this one. Re-reading your initial comment I see it was just  poor accounting while applying this method that was the problem. 

Your mod to GRBL actually sounds like a good permanent fix. It probably should be doing that anyway in view of the unpredictable and possibly catastrophic results when can result. I know GRBL is intentionally tuned to minimal size but this has always struck me as being rather risky. 

I'm guessing this is just a one line change. Could you share that fix, I may add it myself.

bertieconfundo:
It's a small fix. I'm not sure what the difference is to the compiled file.  In my specific configuration it still compiled and fit into the arduino.  Two files affected:

--- a/grbl/serial.c
+++ b/grbl/serial.c
@@ -193,6 +193,17 @@ ISR(SERIAL_RX)
           serial_rx_buffer[serial_rx_buffer_head] = data;
           serial_rx_buffer_head = next_head;
         }
+        else {
+          // The stock firmware silently ignores bytes when the buffer is full. This is dangerous,
+          // and at best leads to invalid/incoherent commands and at worse can lead to a machine crash.
+          // If the serial buffer is ever full when a new byte comes in, the only safe thing to do
+          // is reset the controller and enter an ALARM state. I've created a new alarm code for this
+          // situation which should help in debugging the gcode sender (which in theory should never
+          // overflow the RX buffer)
+          mc_reset();
+          system_set_exec_alarm(EXEC_ALARM_SERIAL_BUFFER_OVERFLOW);
+        }
+        // END TEMP HACK
       }
   }
 }
diff --git a/grbl/system.h b/grbl/system.h
index cfc9273..a9034ad 100644
--- a/grbl/system.h
+++ b/grbl/system.h
@@ -48,6 +48,7 @@
 #define EXEC_ALARM_HOMING_FAIL_PULLOFF        8
 #define EXEC_ALARM_HOMING_FAIL_APPROACH       9
 #define EXEC_ALARM_HOMING_FAIL_DUAL_APPROACH  10
+#define EXEC_ALARM_SERIAL_BUFFER_OVERFLOW     11 // New alarm code to indicate serial buffer overrun - not recognized by senders but still useful

 // Override bit maps. Realtime bitflags to control feed, rapid, spindle, and coolant overrides.
 // Spindle/coolant and feed/rapids are separated into two controlling flag variables.

J-Dunn:
Many thanks. That just two calls and a one byte constant. Looks like 10-20 bytes of object code. There is easily that much slack in  arduino GRBL.  That's worth a PR to GRBL.

PR #1701, opened 2021-10-05
Added support for importing DXF

breiler:
Initial version for importing DXF files in the designer which will make it possible import circuit board designs directly from Kicad. Unfortunately the library Kabeja was the only library I found that supported for reading DXF:s but was not available in maven, so I added the whole library source. @winder decide what you feel about this.

image


winder:
I don't love importing 25k lines of code, but this is pretty cool! Lets go for it :)

PR #1791, opened 2022-01-30
use tar instead of zip/unzip to create windows bundle

arneboe:
the tar binary is intalled by default in windows 10 while zip/unzip need to be installed manually.

### DIFF
--- a/ugs-platform/application/pom.xml
+++ b/ugs-platform/application/pom.xml
@@ -290,10 +290,10 @@
 
                                         <!-- Unzip archive -->
                                         <mkdir dir="${project.build.directory}/${brandingToken}-win/jdk"/>
-                                        <exec dir="${project.build.directory}" executable="unzip" failonerror="true">
-                                            <arg line="-e"/>
+                                        <exec dir="${project.build.directory}" executable="tar" failonerror="true">
+                                            <arg line="-xf"/>
                                             <arg line="jre-win.zip"/>
-                                            <arg line="-d &quot;${brandingToken}-win/jdk&quot;"/>
+                                            <arg line="-C &quot;${brandingToken}-win/jdk&quot;"/>
                                         </exec>
 
                                         <!-- Get the JRE folder name -->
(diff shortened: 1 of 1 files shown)


breiler:
Thanks for the PR.

What do you mean by `zip/unzip need to be installed manually`, from the command line?
This has been possible out of the box since Windows XP and they still have instructions for in the documentation: https://support.microsoft.com/en-us/windows/zip-and-unzip-files-f6dde0a7-0fec-8294-e1d3-703ed85e7ebc

arneboe:
On my windows 10 machine the script failed because zip/unzip where unknown commands. 
zip/unzip in the right-click menu of the explorer works fine. But on the cmdline zip/unzip is not available. tar however is available on the cmdline. Is that not the normal case? :D

breiler:
Ok, you want to be able to compile and create a packaged version of UGS on a Windows-machine using the build scripts. Yes, then you would need install a zip/unzip program on your system to make this work.

I am reluctant to change the packaging format for Windows. I believe that very few are building the software from scratch, instead they are using the prebuilt binaries. With the changes in this PR you will make the installation process more difficult for the majority of people requiring them to have a third party software to unpack the software other than just right clicking on the package and extract as `tar` is only included in Windows 10.

breiler:
Gah, sorry. You are using tar to create a zip-file. 

I was focusing on this error and not the actual changes: https://app.travis-ci.com/github/winder/Universal-G-Code-Sender/builds/245626421#L5547

breiler:
It doesn't look like **GNU tar** has the option to zip/unzip which is used by our build server. 
It works fine on MacOSX using **bsdtar 3.3.2** and apparently on Windows.

arneboe:
Ah damn. I didn't know that GNU tar does not support zip/unzip.

So this fixes the build on windows machines but breaks it on linux machines. And the solution is to either install zip/unzip on windows or bsdtar on linux. Both solutions are not really satisfactory :D

breiler:
Closing as this would break the current build system for us

PR #1797, opened 2022-02-03
Changed to a more obscure separator character for preprocessing files

### LINKED ISSUE #1795: Wrong parsing of comments
### Version

2.0.9

### Hardware / Firmware

Other (add details about it below)

### What happened

If a gcode file contains a comment with the string "++" in it, the parser gives an error and stops processing the file (no preview)

### How to reproduce

Edit a gcode file and add a line like this:
` (++)`
Try to save it.
You get a message box with the following error
`Problem opening file: Corrupt data found while processing gcode stream (++)++++4++++`

### Operating System

Windows 10

### Anything else

_No response_

breiler:
Changed the separator chars "++" to "¶¶". This should be more obscure and would at least fix this case.
@winder can you come up with any reason for not doing this?

Fixes #1795


winder:
Good idea 👍

winder (approved):

PR #1877, opened 2022-05-02
Several fixes to improve stability

hovercraft-github:
UGS just rocks!
However I had to apply the following patches to fix few stability issues:
1) A race condition that manifest itself from time to time in the forEach loop in updateCustomGcodeControls method of MacroActionPanel class is fixed:
```
---------------- On macro run ----------------
INFO [com.willwinder.universalgcodesender.model.GUIBackend]: Sending gcode command:  G10 P0 L20 Z0
SEVERE [com.willwinder.universalgcodesender.GrblController]: Error while processing response <<Run|MPos:-0.038,-0.028,0.000|FS:96,0>>
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1509)
```
2) In the MacroSettingsPanel class the findUniqueMacroName method can
throw a nullException because of empty macro name (fixed). The non-empty name check is not enforced e.g. when we import macros form a json file.
```
---------------- On Add macro ----------------
java.lang.NullPointerException
	at com.willwinder.universalgcodesender.uielements.macros.MacroSettingsPanel.lambda$findUniqueMacroName$11(MacroSettingsPanel.java:356)

```
3) When I accidentally run UGS with any JVM  but aoj jdk-13.0.1+9, an AssertionError is thrown on the job start. The application then continue to work, but its behaviour become weird. In particular the Stop button doesn't work as expected. From the error message given it's hard to understand what's wrong:
```
java.lang.AssertionError
	at org.openide.text.CloneableEditor.getEditorPane(CloneableEditor.java:652)
	at org.netbeans.core.spi.multiview.text.MultiViewEditorElement.getEditorPane(MultiViewEditorElement.java:146)
	at com.willwinder.ugs.nbp.editor.SourceMultiviewElement.UGSEvent(SourceMultiviewElement.java:111)

```
My solution is to catch this exception and give some explanatory message with suggestion to check JVM version. After that the application works normal with usual OpenJDK 13 JVM build, except editing of a G-code is not disabled (looks like a minor problem for me).  
After these small fixes I'm absolutely satisfied with UGS.
Thank you very much for your hard work!

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/uielements/macros/MacroActionPanel.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/uielements/macros/MacroActionPanel.java
@@ -163,7 +163,9 @@ public class MacroActionPanel extends JPanel implements UGSEventListener {
     }
 
     private void updateCustomGcodeControls(boolean enabled) {
-        customGcodeButtons.forEach((button -> button.setEnabled(enabled)));
+        synchronized (this) {
+            customGcodeButtons.forEach((button -> button.setEnabled(enabled)));
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


breiler:
Thanks for digging into this, and thanks for the PR!

PR #1934, opened 2022-07-18
Ethernet feature

villamvadasz:
Hello, we talked some years ago that I had trouble with the TCP and UDP part of the UGS. That time I had trouble to create the PR, but now I did it.
The UDP part I do not use on my CNC machine, since for production it is too unreliable. But was very usefull for supporting development. Makes the uC side implementation much more easier.
The TCP part I use even today actively. Here I had the problem that the 
if(inStream.ready() && (resp = bufIn.readLine()) != null) {
was behaving strange. If in one ethernet frame several lines where recevied, somehow it just received the first line. My guess is (here my capabilities to debug java is very limited) that the ready only becomes true once. And the second line is never received at all. As long as an ethernet based GRBL would send each line in a separate ethernet frame, I guess the old coud would work. But my one was buffering several lines together (I would say this is normally for TCP as a behaviour).
So my solution to this problem was to add a while loop to receive lines as long as there is one.

It looks like to work. But I guess, it would be highly adviced to investigate the problem more deaply before accepting the PR. Again, my capabilities to debug java is limited. I did not had the time to figure it out how to do it.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/connection/ConnectionDriver.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/connection/ConnectionDriver.java
@@ -30,6 +30,7 @@ public enum ConnectionDriver {
     JSERIALCOMM("JSerialComm", "jserialcomm://"),
     JSSC("JSSC", "jssc://"),
     TCP("TCP", "tcp://"),
+    UDP("UDP", "udp://"),
     WS("WebSocket", "ws://");
 
     private final String prettyName;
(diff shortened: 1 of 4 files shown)


breiler:
Thanks for pointing this out. The `TCPConnection.java` actually reads data in a wrong way. It should not attempt to read a line instead it should be reading its bytes. I noticed this when trying to implement the XModem protocol for FluidNC. I am about to do some contract changes to make this more clear.

breiler:
Could you give the latest nightly build a try and see if the TCP connection works as expected with your controller?
I tried it with FluidNC and it looks like it is working.

villamvadasz:
I have checked first the changes in the nightly build. Looks promising, since the reading is now in byte array. And the handler is then taking care of building the string and managing multiple or not complete string. My expectation was that it should work with my machine.
Downloaded and tried out. In deed it was working as expected :)
I see no reason to merge the PR. It can be cancelled or closed because the communication problem is already solved trough an other PR :)

PR #1959, opened 2022-08-11
Fixed macro shortcuts

### LINKED ISSUE #1827: Rename macro breaks its Toolbox button
### Version

2.0.9

### Hardware / Firmware

GRBL 1.1

### What happened

If you have a macro assigned to a Toolbox button, renaming that macro causes glitches in the Toolbox and the macro needs to be re-assigned to a Toolbox button.

### How to reproduce

1) Create a macro
2) Add that macro as a Toolbox button
3) Rename that macro
4) Observe as the Toolbox freaks out; the macro button assumes the name of the button below (above?), then seconds later removes itself.

### Operating System

Win10x64 Pro N

### Anything else

_No response_

### LINKED ISSUE #1862: Macros not updated until UGS is closed & reopened
### Version

2.0.11

### Hardware / Firmware

GRBL 1.1

### What happened

I'm not Gcode expert, but argh! I've been pulling my hair out wondering why the code I've been issuing hasn't worked the way I thought and it's lead me to discovering a bug.

UGS doesn't update macros when you click OK or even if you click Apply first. (Generally OK = Apply + Close.) You need to close UGS and restart it.

### Operating System

Win10x64 Pro N

### LINKED ISSUE #1901: Macros starting with "->" cause disk path error
### Version

2.0.11

### Hardware / Firmware

FluidNC

### What happened

To aid in visual reference, I changed my macro names from "**Go to X0,Y0**" to "**-> X0,Y0**" and in doing so, caused UGS to show an error for each macro.

### How to reproduce

_No response_

### Operating System

_No response_

### Anything else

_No response_

breiler:
Added a UUID to macros in order to identify them uniquely. Previously the name was used to register the macro action which is then stored with that file name. The name may contain illegal characters for the file system which would break any macro shortcut. 

The name of the macro action was not updated which is now also fixed.

Fixes #1901 
Fixes #1862
Fixes #1827

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/types/Macro.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/types/Macro.java
@@ -5,11 +5,13 @@ import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 import java.io.Serializable;
+import java.util.UUID;
 
 /**
  * Created by Phil on 9/6/2015.
  */
 public class Macro implements Serializable {
+    private String uuid = UUID.randomUUID().toString();
     private String name;
     private String description;
     private String gcode;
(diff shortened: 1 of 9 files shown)


daxliniere:
👏👏👏

PR #1968, opened 2022-08-24
fixes a possible bug in JSerialCommConnection.java that, due to re-us…

bertieconfundo:
…e of a common buffer, could result in rx data being lost during overlapping serial event calls

updates jSerialComm to v2.9.2 (has many bug fixes since v2.6.2)

I've only tested this on MacOS.  This PR should be validated against other OS's, as well as trying to reproduce known MacOS jSerialComm hangs to see if the issue is really fixed. (I had a difficult time reliably reproducing the hangs).


bertieconfundo:
closing this PR as it doesn't have the commit I intended ...

PR #2021, opened 2022-10-28
LineSplitter doesn't properly deal with relative coordinate mode (G91)

nickmayer:
GcodePreprocessor was using x start axis for Y & Z
LineSplitter would incorrectly error out on a feed rate only command


nickmayer:
Superceded by request with changes to AutoLeveler

PR #2022, opened 2022-10-29
Updates & fixes for Auto Leveler plugin

nickmayer:
AutoLeveler now warns about possible bugs on applying to gcode instead of spamming on startup
AutoLeveler apply to gcode will now remove old processors and replace instead of disabling button making it possible to use a different heightmap
Fix for MeshLeveler relative mode (G91)


ugs-core/src/com/willwinder/universalgcodesender/gcode/processors/MeshLeveler.java
```diff
@@ -155,17 +153,23 @@ public List<String> processCommand(final String commandString, GcodeState state)
...
+        }
+        if (command.point == null) {
+            return Collections.singletonList(commandString);
+        }
+
+        // If we're in relative mode there is no adjustment needed
+        // TODO: This won't be true if we start in relative mode, in which case we need to ensure the first command
+        // is to move down by the offset at (0, 0)
+        if (!state.inAbsoluteMode) {
+            return Collections.singletonList(commandString);
```
winder, line 167: Is this really true? If there is a difference in the mesh heights between the current point and the previous one doesn't it need to be calculated and applied?

winder: in other words, the current `zPointOffset` minus the previous `zPointOffset`

ugs-platform/ugs-platform-surfacescanner/src/main/java/com/willwinder/ugs/platform/surfacescanner/AutoLevelerTopComponent.java
```diff
@@ -599,26 +599,32 @@ private void dataViewerActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FI
     }//GEN-LAST:event_dataViewerActionPerformed
 
     private void applyToGcodeActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_applyToGcodeActionPerformed
-        GcodeParser gcp = new GcodeParser();
```
winder, line 602: Was something broken with this approach? Without replacing the gcode parser you could have other processors loaded which might effect the final generated program.

ugs-core/src/com/willwinder/universalgcodesender/gcode/processors/MeshLeveler.java
nickmayer: You're right. The previous code was wrong, but so is this. I'll send out an updated patch.

ugs-platform/ugs-platform-surfacescanner/src/main/java/com/willwinder/ugs/platform/surfacescanner/AutoLevelerTopComponent.java
nickmayer: I noticed that `applyGcodeParser` was deprecated and I assumed that having all processors that loaded run would be the expected behavior...wouldn't one expect those other processors to effect the final generated program?

breiler: I remember deprecating this method as I thought it got unintuitive when you applied a processor the previous one got replaced. For instance this prevented the user to apply "Mirror" and "Run from" at the same time.

It is still a bit unintuitive as the changes aren't reflected in the GCode which might have made more sense.

ugs-core/src/com/willwinder/universalgcodesender/gcode/processors/MeshLeveler.java
nickmayer: Updated. I was previously testing with gcode that only had relative z axis-only moves. This time I tested with the gcode below. It does still have an issue when the first move is relative, but I think it requires a broader change to be able properly detect and deal with this situation.

```
G91
G1 Z-1F300
G1 X10Z2F1000
G1 Y10Z2
G1 X-10Z-2
G1 Y-10Z-2
```

ugs-platform/ugs-platform-surfacescanner/src/main/java/com/willwinder/ugs/platform/surfacescanner/AutoLevelerTopComponent.java
nickmayer: +1 to reflecting the changes in the GCode in some way. I considered adding a (read-only) view of what will actually be sent out when I was debugging this.

winder: The default processors would remove comments and I always tried to maintain those in the editor view, but I can see how it's misleading to display different gcode than what gets sent.

daxliniere:
I would be happy to test this new auto-level function if you would like. Just let me know what you need me to do. 👍

nickmayer:
FWIW, I've engraved several PCBs using this code with a 0.05mm depth and it has worked great for me. I do often forget to hit the apply button first though, so I may look into adding a reminder if you have scan data loaded but haven't applied it to your gcode 😉

ugs-platform/ugs-platform-surfacescanner/src/main/java/com/willwinder/ugs/platform/surfacescanner/AutoLevelerTopComponent.java
daxliniere: Would it not be best to apply _all_ processors in a serial fashion, rather than just the last? The order could be pre-determined based on what made the most sense.

daxliniere:
I hate to be 'that guy', but are there directions somewhere on how to build this version into something I can run on my Windows machine, please?

daxliniere (commented):
Wow, fantastic. Nice work Nick and Joacim.
With the recent Github build script addition, does this mean that downloading the Nightly will include this patch?

EDIT: I downloaded the latest nightly (titlebar says 2022-12-28), but I still get the 'Autoleveler doesn't work' warning on startup. @nickmayer indicated [here](https://github.com/winder/Universal-G-Code-Sender/pull/2022#issue-1428373337) that it should only warn when autoleveling is applied to gcode, so I'm left wondering if this nightly is actually up to date.

breiler:
@daxliniere I haven't merged it yet as I have not been able to test it (my workshop is temporarily used for storage  due to some renovations so there is very limited access). You are more than welcome to test it and report any findings.

daxliniere:
Will do Joacim, but won't be back at the workshop for at least another week, I'd say.

daxliniere:
Just extracted it and there's a problem. I got the usual issue that requires cache to be deleted. I deleted it in both locations and on re-loading I got 'module needed, but not found' errors. I clicked Exit, then on re-load again, it loads without error message, but looks like this:

image

There are no cache folders left.

daxliniere:
Uhoh.. looks like the nightly borked my settings. :(
This is how my 2.0.13 looks now:

image

breiler:
@daxliniere this is so weird that this happens to you. I upgrade/downgrade UGS all the time without this ever happening to me.

Anyway, since version 2.0.13 the cache folder is now located under:
`<your user directory/AppData/Roaming/ugsplatform/var/cache`

Sorry for the inconvenience... =(

daxliniere:
Yup, I have deleted both cache dirs, but I still have this problem. Any idea if I can get my layout back?

daxliniere:
> I upgrade/downgrade UGS all the time without this ever happening to me.

How would I perform a clean installation? Maybe that will prevent this happening?

breiler:
> Any idea if I can get my layout back?

When the error occurred it hopefully just broke the window locations. But I am afraid that those settings are lost. As I have never been able to reproduce the error it would have been interesting to see your log file when this happens.

You should be able to restore the window positions by using the menu `Windows / Reset Windows`. It is probably better to try that first, otherwise you will risk losing your joystick mappings if they aren't already gone...

> How would I perform a clean installation? Maybe that will prevent this happening?

A clean install would be to download and unzip UGS:
* Make sure UGS is not running
* Remove all files under C:/Users/[your username]/AppData/Roaming/ugsplatform
* Remove all files under C:/Users/[your username]/.ugs (this is generally never needed)
* Start ugs

PR #2058, opened 2022-11-21
Add FluidNC probe support

nickmayer:
Update FluidNC controller to invoke probeCoordinates on listeners on PRB messages.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCController.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCController.java
@@ -29,11 +29,7 @@ import com.willwinder.universalgcodesender.listeners.ControllerState;
 import com.willwinder.universalgcodesender.listeners.ControllerStatus;
 import com.willwinder.universalgcodesender.listeners.ControllerStatusBuilder;
 import com.willwinder.universalgcodesender.listeners.MessageType;
-import com.willwinder.universalgcodesender.model.Axis;
-import com.willwinder.universalgcodesender.model.CommunicatorState;
-import com.willwinder.universalgcodesender.model.Overrides;
-import com.willwinder.universalgcodesender.model.PartialPosition;
-import com.willwinder.universalgcodesender.model.UnitUtils;
+import com.willwinder.universalgcodesender.model.*;
 import com.willwinder.universalgcodesender.services.MessageService;
 import com.willwinder.universalgcodesender.types.GcodeCommand;
 import com.willwinder.universalgcodesender.utils.IGcodeStreamReader;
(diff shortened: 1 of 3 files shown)


daxliniere:
Is this part of the auto-levelling system?

nickmayer:
This isn't directly part of the auto-leveling system, but is needed by it. Without this the probing would happen but the values wouldn't be reported resulting in it going through the motions but not actually building the heightmap.

IIRC it is also used by some other plugins (I think there is a probe plugin)

daxliniere:
Thanks @nickmayer. Was it you who had recently gotten the autoleveling feature to work in their own fork of UGS?

daxliniere:
Ahh yes, it was you: https://github.com/winder/Universal-G-Code-Sender/pull/2022#issuecomment-1314525542

(I left a question on that post about how to compile this for Windows.)

nickmayer:
> Thanks @nickmayer. Was it you who had recently gotten the autoleveling feature to work in their own fork of UGS?

Yes, I've been using it successfully to carve traces in a PCB using my branch (which I have a PR for https://github.com/winder/Universal-G-Code-Sender/pull/2022). 

There is also https://github.com/nickmayer/Universal-G-Code-Sender/tree/scanner-updates which is a bit less stable. I'm making the auto level data editable (for when the probe doesn't trigger properly and I want to just tweak the one point instead of redoing the entire routine). I am also thinking about tweaking the pattern to go back & forth and allow some other tweaks (like picking the # of points to probe).

I honestly don't know how to build it for Windows. I tried and failed as well, and ended up just building it in an Ubuntu VM. I can take the output from that and run it on my Windows machine (though I had to munge some things in order to get it to find the JVM since it didn't get bundled like the "real" releases do for some reason).

daxliniere:
> There is also https://github.com/nickmayer/Universal-G-Code-Sender/tree/scanner-updates which is a bit less stable. I'm making the auto level data editable (for when the probe doesn't trigger properly and I want to just tweak the one point instead of redoing the entire routine). I am also thinking about tweaking the pattern to go back & forth and allow some other tweaks (like picking the # of points to probe).

You may be interested in knowing about a low-cost, high precision proximity sensor that I've recently learned about. It has 0.01mm accuracy with 0.05mm repeatability and is $37 https://www.pandapi3d.com/product-page/bed-distance-sensor
I'm talking with Mark of PandaPi at the moment and should have one in the new year.

daxliniere:
> I honestly don't know how to build it for Windows. I tried and failed as well, and ended up just building it in an Ubuntu VM. I can take the output from that and run it on my Windows machine (though I had to munge some things in order to get it to find the JVM since it didn't get bundled like the "real" releases do for some reason).

Hopefully @winder or @breiler can point us in the direction of a tutorial for jFrog or some other means of compiling?

daxliniere:
@nickmayer, is there a way to send you a PM? Are you on the FluidNC Discord server, by any chance?
I have something I'd like to share with you.

nickmayer:
DarkStranger#5020 on Discord

Tigeruno:
I have FluidNC in use for my Rong Fu 30 mill.  The resistance from quill/tool to vice/table is less than 2 ohms so probe  with UCS (universal code sender) did not work.  I just found a post with EEF (electronic edge finder by R, G. Sparber) https://vdocuments.mx/a-precision-electronic-cutter-touch-down-detector.html.  I built the last schematic in this link using the leds.  I found that if you connect 5v from esp32 for power and gnd, and instead of LEDS connect probe leads to LM324 pin 14 to 1k resistor to gnd on esp32, LM324 pin 8 to probe pin (for me gpio32 low pullup).  This EEF circuit can detect touch below 2 ohms resistance.  Works a charm for me.

PR #2123, opened 2023-01-15
Add action for following the running gcode in the editor

breiler:
It will select the line that was last completed by the controller when using the toggle button in the editor.


chrisla23:
Given that this is worked out, could this line info be added to a popup
when it goes into an alarm state?

Thanks,

-Chris

breiler:
@hrenz it should be available here in version 2.0.14 or later:
image

This feature is completely new, did not want to turn it on by default as I don't know if it has performance implications.

hrenz:
no, no field no button

breiler:
> 2.0.14 or later

You are running 2.0.12, dude please stop wasting my time...

hrenz:
sorry now its 2.0.14

Issue #1312, opened 2019-11-18
Center Probe Code


dfox1787:
Hi guys

Does anyone have any code to find the center of a Hole? Im using a probe that i want to use to find a hole center?

![Capture3](https://user-images.githubusercontent.com/28814095/69095930-3cf90880-0a4b-11ea-984a-8a0f9b356ce0.PNG)

breiler:
You could do this manually by using the expressions feature in the DRO:
http://winder.github.io/ugs_website/guide/platform/#digital-read-out

1. Go to the left of the hole, reset X
2. Go to the right of the hole, type "/2" in the field for the X work coordinate in the DRO 
3. Go to the X zero
4. Go to the bottom of the hole, reset Y
5. Go to the top of the hole, type "/2" in the field for the Y work coordinate in the DRO
6. Go to the Y zero

This should be the center, but you could run the steps one more time for it to be exact.
There's a video link in #977 that describes this process.

dfox1787:
thanks. im looking to add this in as a macro if possible.

owenduffy:
I think that enhancement of the probe module to exploit the type of calibration fixture shown would be useful. It the plate is made to sufficient accuracy, the calibration is independent of the cutter size (using the small hole which is centred over the work zero).

risototh:
I'm currently working on this. I have 80% of code completed, just strugling with the OpenGL as I want to visualize the process - jsut show a block with the hole and arrows... Arrows are not a problem, but the solid block with hole seems to be a problem for me, for now, today :) But I'm prettu sure, that I will break it. Here is a preview https://youtu.be/gPf3ezeDDHY

owenduffy:
Hi,

I looked at your video, and it looks promising.

Might I suggest you consider the case where the hole is not much larger than the cutter diameter (eg a 5mm hole and 3mm cutter).

Accuracy might be improved if you:

1. probe X- to X+ to find the X centre;
2. move X to centre;
3. probe Y1 to Y+ to find Y centre;
4. move Y to centre;
5. now redo X (now the cutter is approximately centred);
6. now redo Y (now the cutter is approximately centred);

Are you considering including a Z touch off in the same process? You might set the cutter / probe above the Z surface and over the hole, lower the cutter by a specified amount to probe the hole, return to initial Z, move a given X,Y to perform the Z touch off.

Thanks
Owen

risototh:
Hi Owen,

[...]

PR #2197, opened 2023-04-10
Feature/hole probing

risototh:
Probing of hole center issue #1312 
Some translations to Slovak.


breiler:
Nice work, I will give this a closer look and try it out later tonight.

risototh:
I have also reversed idea - center of a stem... Maybe next time I will implement it too, but now I'm heading to the manual tool changes with probing of TLO... This is something I really need (and I'm sure, that not only me...) and should be supported in the grblhal. Now waiting for the TLO probe...

ugs-platform/ProbeModule/src/main/java/com/willwinder/ugs/platform/probe/ProbeService.java
```diff
@@ -378,6 +381,91 @@ private void performXYZProbeInternal(int stepNumber) throws IllegalStateExceptio
...
+        try {
+            switch (stepNumber) { // NOTE: G code comments are with radius 25 and retract 2, G21, G54
+                case 0: { // find -X
+                    // Reset (0,0,_) to make it easier to retract.
+                    updateWCS(params.wcsToUpdate, 0.0, 0.0, null); // G10 L20 P0 X0 Y0
+                    probe('X', params.feedRate, -holeRadius, params.units); // G21 G91 G49; G38.2 X-25.0 F250
+                    break;
+                }
+                case 1: { // retract & measure -X
+                    gcode(g0Rel + " X" + params.retractAmount); // G91 G21 G0 X2.0
```
breiler, line 409: I can see that we previously did it this way, just concatinating strings and doubles.

The problem is that this is locale sensitive. In some locales a minus character is UTF-8 encoded using two bytes. Commas may also be represented with a `,` instead of a `.`.

Can you replace your concatinations of doubles using `Utils.formatter.format(x)` or similar instead now that you have code fresh?

risototh: Thanks for review, forgot the locales stuff... Should be fixed now.

PR #2200, opened 2023-04-16
Smart scan

nickmayer:
Reworked the auto leveler to be a lot smarter about scanning resulting in a huge speedup. Also change UI to make it a lot less likely you'll forget to apply your scanned heightmap to your gcode.


nickmayer:
I see there are some tests failing...not too surprising since I can't actually run the tests with my setup (MSYS on Win10). What is the "official" setup used by the devs? I haven't had any luck getting `mvn` to work without skipping tests (`mvn install -Dmaven.test.skip=true`) on either Win10 MSYS, Win10 WSL2, or an Ubuntu VM.

breiler:
Sorry for the lack of response on this. I'm currently rebuilding my machine and have ordered a decent probe to test this out. I'll take care of the merge conflict.

breiler:
I have now finally tried this and it seems to work fine. I needed to revert the changes to the `CommandProcessorList` to not break its behavior described in my previous comment.

There was also a place where you added a double value comparison. This did in certain cases not work because double values are not exact or that there could be rounding errors reported from the controller. I had to add a double compare function with a minimum error threshold.

I was thinking of creating a PR to your branch, but it got pretty messy with all changes coming from master. So I'll close this and create a new PR.

Thanks for your contribution!

PR #2218, opened 2023-05-06
Parse version string from millright firmware

breiler:


bLouChip:
So I have mild USB connection issues using UGS v2.0.12 in that I have to disconnect/connect several times before the connection is made as evidenced by console listing grbl version and $$ report.  That's not really a problem for me, just curious if this fix was intended to solve the need for multiple retries.  I'm using MillRight MegaV CNC w Win 11 and JSerialComm config.  

But more importantly, I have modified MillRight's grbl from their source and thus the version and build signature, running on the stock MillRight mega2560 board.  This fix in UGS seems to suggest it's looking for the exact original grbl version and build signature by MillRight.  Why is this being done ?  Does it inhibit the connection to my grbl ? 

Here's my grbl version and signature: 
"Grbl 1.1i MR MegaV bLouChip Tri-CAM Open Frame Bed ['$' for help]"
and here is the $i report: 
">>> $i
[VER:1.1i MR MegaV bLouChip Tri-CAM Open Frame Bed.V4-20230217:]
[AXS:4:XYZA]
[OPT:VNMPHRL,35,255]"

breiler:
This will apparently be a future problem...

We need to get the controller version to figure out what protocol to use to speak with the controller. Previously this was done using the welcome string:
```Grbl 1.1i MR MegaV bLouChip Tri-CAM Open Frame Bed ['$' for help]```

In order to be able to communicate with newer variants of GRBL firmwares, the welcome string could no longer be used for this. So I opted for using the `$I` command and decided to try and get the machine name and build date for future features. But the specification is apparently too loose which makes it impossible to parse this information...

The exact version string that you see in this PR is a test for a known variant of a version string. So far I have needed to parse ten different structures of how this string can be written. Yours makes number eleven.

I will simplify this to just parse the version string `<major>.<minor><character>` such as in `1.1h`. There are enough variants of it to try and cover:

[VER:1.1i MR MegaV bLouChip Tri-CAM Open Frame Bed.V4-20230217:]
[VER:v1.1f.20170131:Some string]
[VER:1.1i MegaV 4 Axis Router.20190120:]
[VER:1.1g3.20211002.Mega:]
[VER:1.1f.20170801:MINIMILL]
[VER:1.1h-XCP.20220314a:abc]
[VER:v1.1.20170131:Some string]
[VER:0.7]

breiler:
@bLouChip please try the latest nightly build:
https://github.com/winder/Universal-G-Code-Sender#downloads

breiler:
That is correct, both version strings should be detectable by UGS. If others still have problems they can post a new bug report and I can have a look.

Issue #1468, opened 2020-11-08
Request: Support for grblHAL


phil-barrett:
grblHAL is a multi-platform 32 bit version of Grbl.  It adds new features like canned cycle, tool change and a number of control and reporting options as well as things like input button pin invert as well as things.  It also supports ethernet and wifi connectivity.

Currently, grblHAL needs to be compiled to run in compatibility mode to work with UGS but in doing so it omits the new features and control options. It would be great if UGS was able to work directly with grblHAL native. Basically, I believe the issue is handling the additional messages that grblHAL generates.

grblHAL currently runs on 13 different microcontroller platforms.  I have a breakout board for grblHAL based on the Teensy 4.1 that I sell on [Tindy](https://www.tindie.com/products/philba/grblhal-breakout-board-unkit-for-teensy-41/). Currently I don't recommend UGS to my customers but would like to be able to.  This is especially true for Linux and Mac based platforms. I'd be happy to help with any questions.  The [github repository for grblHAL is here](https://github.com/terjeio/grblHAL) and the issues section is quite active. TerjeIo, the developer, is quite helpful and, I am certain, would be more than happy to help. There is even a thread about using UGS.

breiler:
Your break out board looks really nice, I think that I will order one for myself for Christmas. 😄 
I took the liberty to add a few MVP-features checklist to help us keep track of the support implementation progress.

phil-barrett:
Thanks.  I have one customer that is rather pointed about wanting UGS and nothing else! He would be over the moon.

I think it should be relatively easy to add support though those are famous last words more often than not!

Contact me through tindie and we can work something out.

Phil

andrewfernie:
I saw the commit for basic GrblHAL support so tried the nightly. No problem getting it going on Windows with the latest GrblHAL running on a Teensy 4.1. 

One minor issue is trying to use a keyboard shortcut to run the homing routine. I assigned Alt-F6 to home so that I could run it from a programmable keyboard. This works fine with a standard grbl controller but not with GrblHAL.  Doesn't matter if I use my programmable keyboard or typing Alt-F6 on the standard PC keyboard, there is no response. Unplug the controller and plug in a standard grbl controller and it works. All my other keyboard shortcuts work fine with either controller. 

No problem running home with the UGS UI button, so not a big deal, just strange that home doesn't work as a shortcut.

breiler:
[...]

PR #2275, opened 2023-08-09
Fixed a problem with grblHAL

breiler:
Fixed a problem with grblHAL where certain responses were treated as parser state commands.


tcurdt:
@breiler I am confused - should UGS already work with grblHAL? At least the tool change mode seems to break things for me https://github.com/winder/Universal-G-Code-Sender/issues/1468#issuecomment-1694646153

breiler:
@tcurdt the things that are compatible with GRBL will work in UGS. Tool change for instance does not exist in GRBL, so that may very well cause problems.

PR #2404, opened 2023-12-27
Enhancement: Advanced macros

andrewmurraydavid:
I updated the macros to support multi-line macros, variables and basic math operations (`+, -, /, *`).
With this in place, more advanced feature could be added, like basic functions (`ROUND`, `FIX`, etc.) in the future.

Should be backwards compatible with existing macros (single-line macros).

I need this for more advanced macros including probing, slot/hole drilling for fixtures (things that need to be repeatable exactly the same without copy-pasting code).

⚠️ **MAKE A BACKUP OF YOUR MACROS BEFORE TESTING THIS LOCALLY**, I'm still working on a fix for this.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/MacroHelper.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/MacroHelper.java
@@ -22,29 +22,17 @@ import com.willwinder.universalgcodesender.i18n.Localization;
 import com.willwinder.universalgcodesender.model.BackendAPI;
 import com.willwinder.universalgcodesender.model.Position;
 import com.willwinder.universalgcodesender.services.KeyboardService;
-import com.willwinder.universalgcodesender.utils.GUIHelpers;
 import net.miginfocom.swing.MigLayout;
 import org.apache.commons.lang3.RegExUtils;
 import org.apache.commons.lang3.StringUtils;
 
-import javax.swing.JLabel;
-import javax.swing.JOptionPane;
-import javax.swing.JPanel;
-import javax.swing.JTextField;
-import javax.swing.KeyStroke;
-import java.awt.AWTException;
-import java.awt.Robot;
-import java.awt.event.InputEvent;
+import javax.swing.*;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
-import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
-import static java.awt.event.KeyEvent.VK_CONTROL;
-import static java.awt.event.KeyEvent.VK_META;
-import static java.awt.event.KeyEvent.VK_SHIFT;
 
 /**
  * Static helper class for executing custom macros.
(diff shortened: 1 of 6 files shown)


andrewmurraydavid:
@winder @breiler Java isn't the language I have the experience most in, so let me know what needs fixing or improving!

breiler:
Thanks for the PR, I am currently travelling so I can't review this at the moment.

ugs-core/src/com/willwinder/universalgcodesender/MacroHelper.java
```diff
@@ -79,13 +82,14 @@ public class MacroHelper {
...
      * @param backend
      */
-    public static void executeCustomGcode(final String str, BackendAPI backend) throws Exception {
+    public static void executeCustomGcode(final String[] str, BackendAPI backend) throws Exception {
         if (str == null) {
             return;
         }
-        String command = MacroHelper.substituteValues(str, backend);
-        command = command.replaceAll("(\\r\\n|\\n\\r|\\r|\\n)", "");
-        final String[] parts = command.split(";");
```
breiler, line 88: Macros does support multiline today with a `;`-separator. It looks like you've removed that behaviour which will impact existing macros.

I think it is better to use `\n` instead of `;` like you have suggested. But instead of just dropping the old behaviour you must do some sort of conversion. As `;` is commonly used as a comment prefix, this new conversion should probably only be done once.

May I suggest adding a version to the `Macro`-class. If it doesn't exiss you should treat `;` as newlines and convert. If it has a version tag it should expect `;` as gcode comments and `\n` as newlines.

And by doing this, you don't have to change the type from `String gcode` to `String[] gcode`. And you probably won't need the MacroSerializer and MacroDeserializer.

ugs-core/src/com/willwinder/universalgcodesender/MacroHelper.java
```diff
@@ -22,30 +22,18 @@ This file is part of Universal Gcode Sender (UGS).
...
 
-import javax.swing.JLabel;
-import javax.swing.JOptionPane;
-import javax.swing.JPanel;
-import javax.swing.JTextField;
-import javax.swing.KeyStroke;
-import java.awt.AWTException;
-import java.awt.Robot;
-import java.awt.event.InputEvent;
+import javax.swing.*;
```
breiler, line 29: This is a minor thing. Can you change the settings in your IDE to not import *.
This gets annoying when for instance I want to import `java.util.List` but instead get a `javax.awt.List`.

breiler:
Could you split the PR into two - one for the multiline support and then one for the scripting?

I have been keeping an eye on @asmr-hex branch for adding scripting to UGS which I think is the way forward. The implementation is based on the javascript engine:
https://github.com/winder/Universal-G-Code-Sender/compare/master...asmr-hex:Universal-G-Code-Sender:expression-support

We already us it for some simple scripting in the DRO here:
https://github.com/winder/Universal-G-Code-Sender/blob/master/ugs-core/src/com/willwinder/universalgcodesender/model/GUIBackend.java#L725

Your method could be a start, but will probably be replaced with the JavaScript engine later on.

You are using `[]` as a placeholder for scripts, which conforms with cncjs. But we will probably not be 100% compatible if we go the JavaScript route. @asmr-hex is using `{}` which I also like as it resembles the javascript scopes. Both have its benefits, so if anyone have any input, please share.

andrewmurraydavid:
Thanks for your initial review, @breiler !

Replacing with the JS engine is no problem, I wasn't aware that there's already work on using expressions.
Happy to replace `[]` with `{}` for future-proofing. I used LinuxCNC and it uses `[]` for expressions, but them other interpreters use `{}`, so it's more of a "business" decision.

I'll focus this PR on multiline and create a new PR for advanced macros, if that works for you, or I can close this and create separate PRs.

PR #2436, opened 2024-01-23
Add Driver selector to Connection setup wizard

### LINKED ISSUE #2425: Proposal to enhance Setup wizard for FluidNC
### Current behavior

Currently, UGS assumes that the connection driver has already been set correctly when selecting the firmware and connection settings. When UGS essentially only supported serial connections because most controllers were connected using a local serial connection, that was a reasonable assumption.

### What problem does this feature solve?

Newer controllers can be controlled over the network, either directly with TCP or using Websockets. FluidNC is an example of a newer CNC controller that is overwhelmingly accessed over the network.

The FluidNC maintainers explicitly tell users to use more full-featured senders like UGS so they can concentrate on just connecting to and controlling the CNC machine itself.

When a new user is presented with the connection options in the Setup wizard as it exists today, and they want to connect to a networked controller, it's not clear that they need to change the connection driver from the default `JserialComm` driver to `TCP`.

Once the user figures out they need to change the driver, it's a relatively high number of UI actions to close out the Setup wizard, open the Connection panel,  change the driver, restart the Setup wizard, and then attempt the connection.

This can be simplified and provide a better initial user experience (and with fewer obnoxious support requests) by adding additional behavior to the Setup wizard.

### Proposed behavior

The Connection setup wizard should allow the user to select the connection driver from within the wizard. Users can then choose how they will connect to the controller within a single UI panel. Users can switch back and forth from within the wizard for controllers that can connect through both serial or network connections.

### Initial proposal (abandoned)

A Setup wizard can be more opinionated about connection options and default values.

The Setup wizard should make reasonable assumptions about the controller to which is connects. If e.g. the user selects FluidNC or Smoothieware, it is reasonable to assume that the type of connection is over the network, so the connection type should be set to a `TCP` from within the wizard itself. The UI can inform the user that selecting FluidNC or Smoothieware will change the default driver, and allow the user to cancel out if that's not desired.

After the user selects e.g. FluidNC, and the input boxes are updated to reflect the options required for TCP connections, the options should be prepopulated with any appropriate values for that firmware. In the case of FluidNC, the default port for TCP connections is the Telnet port, 23.

### Alternatives considered

The Setup wizard could alert the user that the preferred driver for a given firmware choice is not currently active. This would point users towards the correct action of changing the driver, as right now there's no feedback from UGS that they are on the wrong path. This would be simpler to implement. The downside is that the user would have to take additional action by manually closing out the Setup wizard, opening the Options dialog, navigating to the Sender Options pane, selecting the correct driver, applying the changes, closing the Options dialog, then restarting the Setup wizard.

**Edit:** This alternative was chosen as correct solution.
The Setup wizard could add a driver combo box so users could select the driver from within the wizard before choosing the firmware. This again is a simpler option to implement than the proposal. The downside is that users might not understand which driver is the preferred driver for a given firmware choice, and just move on to selecting the firmware while leaving the current driver unchanged. This will result in connection errors, and probable user frustration.

### Goals and non-goals of this feature

Goals:

- Add the ability to dynamically change the connection driver based on the firmware type to the Setup wizard.
- Provide reasonable defaults to the user in the Setup wizard based on the selected firmware.
- Provide a generalized framework for giving configuration hints to firmware types.

Non-goals:

- Refactor the Setup wizard.
- Validate user provided configuration options.
- Troubleshoot connection problems within the Setup wizard.

shampeon:
This PR adds a combo box for selecting the Driver in the Connection setup wizard.

Fixes #2425. 

To do:
- [x] Fix layout issue with Connect button being hidden.
- [x] Move labels/descriptions to localization bundle.

### DIFF
--- a/ugs-platform/ugs-platform-plugin-setup-wizard/src/main/java/com/willwinder/ugs/nbp/setupwizard/components/ConnectionPanel.java
+++ b/ugs-platform/ugs-platform-plugin-setup-wizard/src/main/java/com/willwinder/ugs/nbp/setupwizard/components/ConnectionPanel.java
@@ -51,13 +51,14 @@ public class ConnectionPanel extends JPanel implements UGSEventListener {
         // Driver options
         driverCombo = new JComboBox<>(ConnectionDriver.getPrettyNames());
         driverCombo.addActionListener(d -> this.setDriver());
-        JLabel labelDriver = new JLabel("Driver:");
-        labelDriver.setToolTipText("Select the driver to use to connect to your firmware. Serial connections should use JSerialComm. Network connections should use TCP.");
+        JLabel labelDriver = new JLabel(Localization.getString("settings.connectionDriver"));
+        labelDriver.setToolTipText("Select the driver to use to connect to your controller firmware. Serial connections should use JSerialComm. Network connections should use TCP.");
 
         // Firmware options
         firmwareCombo = new JComboBox<>();
         firmwareCombo.addActionListener(a -> setFirmware());
-        JLabel labelFirmware = new JLabel("Firmware:");
+        JLabel labelFirmware = new JLabel(Localization.getString("mainWindow.swing.firmwareLabel"));
+        labelFirmware.setToolTipText("Select the controller firmware to which you want to connect.");
 
         // Baud rate options
         baudCombo = new BaudComboBox(backend);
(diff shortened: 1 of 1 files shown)


shampeon:
I'm seeing a layout problem where the Connect button is pushed to be barely visible, but no scroll bars appear.
Screen Shot 2024-01-22 at 5 52 38 PM

Ideally, the Setup Wizard panel would expand on the y axis to fit the full `ConnectionPanel` elements.

breiler:
Thanks

PR #2467, opened 2024-02-16
Some small designer fixes 

werckmeister:
This pull request includes some minor adjustments aimed at making it easier and more natural to create and position objects in the designer.

1. Now, when you create a basic shape like an ellipse, point, or rectangle, it will be automatically selected for you.
2. A move action (and related events) will only occur if there is a noticeable change (delta > 0). This prevents accidental moves when clicking on certain controls.
3. The bounds (size and position) will be updated immediately after a change but before any notifications are sent out. This ensures that event handlers receive the most up-to-date information and don't use outdated cached data.


breiler:
Feels much nicer, thanks!

Issue #2473, opened 2024-03-01
Automatically focus macro placeholder input


warmans:
### Description

I use one of those wireless numpads to control UGS e.g.

![image](https://github.com/winder/Universal-G-Code-Sender/assets/1299401/ff16ad5c-51c5-4241-8bcd-2b9f5a59083d)

This works great with macros as it's easy to just bind them to a function key, and use the arrows for jogging etc. Typically I use macros with numeric `{{placeholders}}` (e.g. the tool diameter). So what I would like to do is:

1. press the macro key (e.g. X probe)
2. enter the value with the numpad 
3. hit enter 

However there is one small problem: the input box of the macro pop-up is not focussed when it opens. So you still need to grab the mouse and click in the input box. 

**My request is that when a macro placeholder popup opens, it will automatically focus on the input. 🙏**

PR #2474, opened 2024-03-02
fix: Set focus to first field in macro popup

warmans:
Resolves: https://github.com/winder/Universal-G-Code-Sender/issues/2473

I've not touched Java in many years so this solution is taken mostly from here: 

https://tips4java.wordpress.com/2010/03/14/dialog-focus/

It seems to behave differently on different platforms. HierarchyListener was the only listener that worked on Linux.


breiler:
Nice, thanks!

Can you move the `RequestFocusListener` to the package `com.willwinder.universalgcodesender.uielements.components` and also add a comment that credits the original source, something like:

```java
/**
 * Used for focusing the field in an JOptionDialog. 
 * Source from: https://tips4java.wordpress.com/2010/03/14/dialog-focus/
 */
public class RequestFocusListener implements HierarchyListener {
...
```

PR #2475, opened 2024-03-02
Support default value in macro prompt

warmans:
1. Adds an optional third value to prompts that contains the default value e.g. `{prompt|pos|0}` would show `0` in the input labeled `pos` when the panel opens. 
2. Updates help text to describe how to set defaults (also adds note about reusing placeholder values).
3. Removes some unused imports.


breiler:
Clean and simple. Can you just add a couple of `<br/>` in the the helper text as it gets quite long at the moment:

image

warmans:
Hi, thanks for reviewing. I've tidied up the text a bit and put in some breaks. Let me know if there are any other changes needed.

image

PR #2489, opened 2024-03-20
Added rapid speed overrides

breiler:
Added rapid override buttons for GRBL:
image

Also made the component able to become compact:
image


bLouChip:
Sounds great.  I'll give it a try on the next overnight build.  
Thanks much.

bLouChip:
well done.  The override window scales to size nicely, and the rapid override controls are present as well. 
The Spindle, Mist, and Coolant buttons are still only shades of gray on my Win 11, rather than previous UGS versions of green and red depending on state. 
Also, the override grbl cmds are now suppressed from the console log, making it hard for me to know the command state if I can't rely on color of button.  I will give it a good shake down with the CNC tomorrow, that's a Win 10 computer.

bLouChip:
@breiler ,  the override window adjustments you made are working great.  overall window size scaling is unlimited it appears, as are other windows now, with auto adjust font size.  Well done.  Rapid override working fine too.  I like the new file dialogue UI using system template. gcode editor seems more stable now, working fine;  not locking up the process when keeping focus while running a job.  Visualizer white control point pointer is now in sync with system mouse pointer when system screen is scaled (Win10-11).

Issue #2673, opened 2025-01-09
ugs probe is not in the expected initial state


bvernham:
### Version

2.1.9

### Hardware / Firmware

GRBL 1.1

### What happened

I keep getting this error with UGS "probe is not in the expected initial state" which is very confusing becuase I have never had and issue with probing before.  I have probed with bCNC and with EASEL with zero issues.  With UGS I am getting (ALARM:4) Probe fail. Probe is not in the expected initial state before starting probe cycle when G38.2 and G38.3 is not triggered and G38.4 and G38.5 is triggered.  The machine is zeroed with everything hooked up (clip to the bit) and to the touch plate.  The machine probes down to the touch plate.  Makes contact, backs off, then set the alarm.  Probe cycle looks OK until the makes contact (like it should) then it fails?  Never had thie problem with any there g code sender.

### How to reproduce

See above.  Tried multiple times, same issue.

### Operating System

Windows 10

### Anything else

Every time I probe.  I wouldn't make this an issue if I had no probing experience with other applications.  But with Easel you can use machine inspector and the machine check out as expected.  The limit switches only set when you push them (as expected) and the probe is registered only when the touch plate is manually touching the bit.

bvernham:
>>> G10 L20 P1 Z0
>>> 
>>> G21 G91 G49
>>> G38.2 Z-90 F100







>>> 
ok
ok
ok
[PRB:0.000,0.000,-82.823:1]
>>> G91 G21 G0 Z2.0
>>> 
>>> G21 G91 G49
>>> G38.2 Z-90 F10
>>> 
ok
ok
ok
ok
ok
**** The communicator has been paused ****
An error was detected while sending 'G38.2 Z-90 F10': (ALARM:4) Probe fail. Probe is not in the expected initial state before starting probe cycle when G38.2 and G38.3 is not triggered and G38.4 and G38.5 is triggered. Streaming has been paused.
ok
Error while processing response <ok>: An unexpected command was completed by the controller.

I tried a bigger #....

bvernham:
Now I am getting this:

>>> G10 L20 P1 Z0
>>> 
>>> G21 G91 G49
>>> G38.2 Z-90 F100
>>> 
ok
ok
ok
[PRB:0.000,0.000,-82.844:1]
ok
ok
>>> G91 G21 G0 Z3.0
>>> 
>>> G21 G91 G49
>>> G38.2 Z-90 F10
>>> 
ok
ok
ok
[...]

PR #2702, opened 2025-02-27
Add probe "delay after retract" setting

paulstraw:
For #2673. Adds a "delay after retract" setting to the probe module. I tested Z, XY, XYZ, and Hole center on my machine.

I assume I may have missed a few things since this is my first contribution. Feel free to adjust, or let me know what needs to be changed!


breiler:
Merged in #2715

Issue #1491, opened 2020-12-12
Configurable Minimum DRO Resolution and Jog Step Size


bgoodman44:
DRO display resolution and minimum jog step size are not user configurable.  They should be.  Currently, DRO display is fixed at 3 decimal places and jog size is limited to 0.001"

PR #2756, opened 2025-04-17
Initial Implementation DRO Resolution

dnikodem84:
- Added concept of "machine precision" to control Decimal Places in DRO and Jog Spinners
- Added new PopupMenu Sub Menu to allow user to change decimal places.
- Updated DRO Controls to update their decimal places
- Updated StepSizeSpinner control's to all use

This is based on: https://github.com/winder/Universal-G-Code-Sender/issues/1491


breiler:
What controller are you using that supports more than three decimals?

dnikodem84:
Even with a Step size of 800 per mm you are unlikely to need that precision, when dealing in metric.

When converting from Inches to MM and back then there is a rounding error which can change the values and this is what I am concerned about here. 

5 decimal places is enough to have 1mm convert successfully to inches and back.

breiler:
Too me as that is used to work in metric this got really confusing.

The smallest unit of change that the controller can report is 0.001 mm. So with the option to have five decimals will guarantee a lot of support questions from metric users. I don't think that this option should be available when using metric.

Having the option to add more precision when using imperial units makes sense. 
The prerequisite must however be that the controller is set to report in metric which not all controllers are. Otherwise this will cause confusion to imperial users too as they wouldn't see any changes below three decimals in the DRO.

The jog command generation needs to be changed to allow more precision. Otherwise the length of the jog command will become zero for small imperial jogging. It could be as easy as changing the the `Utils.formatter` here:
https://github.com/winder/Universal-G-Code-Sender/blob/master/ugs-core/src/com/willwinder/universalgcodesender/gcode/util/GcodeUtils.java#L63C45-L63C54

The controllers that I know of does not support decimal feed rates, so I suggest that the rate field only shows as an integer. The step size range should also allow stepping down to the number of decimals:
image

I have some concerns about making the StepSize-spinner dependant on the BackendAPI. I'll write them up in some other comments, I am however out of time... =(

dnikodem84:
I decided to put the new options in the Backend since the "seperate step sizes" option was already there and it felt like other  view only options could be join it. 

I had to have a crash course in imperial after moving to America, and honestly it feels like having "x / y " style of input field which maintains the fractions is the best option for imperial. ( 1 / 4 , 1 / 64 are both more readable than 0.25 and 0.015625 ) 

All of the review comments seem sensible and I will have some time to work on this tomorrow after work so please let me know of any other review comments you may have.

ugs-core/src/com/willwinder/universalgcodesender/uielements/jog/StepSizeSpinner.java
breiler, line 1: Is it possible to keep this component simple without the dependency to the backend? The component gets a bit "heavy" as it is now. 

Not sure where the best place to put this setting, there is a place where these settings are read and updated today. The downside is that the setting needs to be "prop drilled" all the way down to the components that needs them. But we get rid of the rather large dependency to the backend:

https://github.com/winder/Universal-G-Code-Sender/blob/master/ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/JogTopComponent.java#L133

ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/actions/SetMachineDecimalFormatFour.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public JMenuItem getMenuPresenter() {
+        return menuItem;
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        backend.getSettings().setMachineDecimalFormat(DECIMAL_FORMAT_FOUR) ;
+        backend.getSettings().changed();
```
breiler, line 101: Move this call to changed() into the Settings::setMachineDecimalFormat, then you can hide the changed method again to private.

ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/actions/ShowABCStepSizeAction.java
```diff
@@ -93,5 +94,6 @@ public JMenuItem getMenuPresenter() {
     public void actionPerformed(ActionEvent e) {
         // Toggle the usage separate Z step size
         backend.getSettings().setShowABCStepSize(!backend.getSettings().showABCStepSize());
+        backend.getSettings().changed();
```
breiler, line 97: This should not be needed as "setShowABCStepSize" already calls this after the update.

ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/actions/ShowABCStepSizeAction.java
```diff
@@ -69,6 +69,7 @@ public ShowABCStepSizeAction() {
         backend = CentralLookup.getDefault().lookup(BackendAPI.class);
         backend.addUGSEventListener(this::onBackendEvent);
         setEnabled(isEnabled());
+        onBackendEvent( new SettingChangedEvent()) ;
```
breiler, line 72: Good find! It is a bit obscure that this means initialize with the current value. You can leave this as is, but I would probably extract the `menuItem.setSelected(backend.getSettings().showABCStepSize());` to a method `updateSelected` or something.

ugs-core/src/com/willwinder/universalgcodesender/uielements/jog/StepSizeSpinner.java
breiler: > If the "decimalPlaces" property changes the behavior of the output to the controller after a G20 has been sent then I feel as if it does belong to the backend as opposed to global settings.

It does not change the output, but rather jog controller doesn't handle jog movements with numbers smaller than three decimals. That should be an easy fix which can be done at a later time.

The bigger problem as I see it is that the controller may be set to [report its position in inches](https://github.com/gnea/grbl/wiki/Grbl-v1.1-Configuration#13---report-inches-boolean) (the default is metric) or if the user is using metric then these new options will not work well at all. But since these new options are quite hidden we can ignore that for now.

So the only thing left is the dependency to the backend. I'd prefer if you tweak this to instead use the settings. 

I would personally have expanded these places for that:
https://github.com/winder/Universal-G-Code-Sender/blob/master/ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/JogTopComponent.java#L133
https://github.com/winder/Universal-G-Code-Sender/blob/master/ugs-platform/ugs-platform-plugin-dro/src/main/java/com/willwinder/ugs/nbp/dro/panels/MachineStatusPanel.java#L223

ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/actions/ShowABCStepSizeAction.java
dnikodem84: Applied in branch

ugs-platform/ugs-platform-plugin-jog/src/main/java/com/willwinder/ugs/nbp/jog/actions/ShowABCStepSizeAction.java
dnikodem84: Fixed in branch.

ugs-core/src/com/willwinder/universalgcodesender/uielements/jog/StepSizeSpinner.java
dnikodem84: - Updated Utils.formatter on a branch
- Changed dependency on BE to Settings.
- Leaving Review comment open for now.

ugs-platform/ugs-platform-plugin-jog/src/test/java/com/willwinder/ugs/nbp/jog/JogPanelTest.java
```diff
@@ -29,7 +29,7 @@ public static void main(String[] args) throws Exception {
     }
 
     private void start() throws Exception {
-        jogPanel = new JogPanel();
+        jogPanel = new JogPanel(null);
```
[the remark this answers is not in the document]
dnikodem84: Currently have test case failures, but must stop for now. I will address this issue when I address those.

PR #2843, opened 2025-07-27
adding a text field for the spindle speed in the 'Cut settings' dialog

mchristianl:
Hi, I would suggest adding add a text field for the laser power (spindle speed) to allow entering low values (e.g. 4%) for laser engraving: 

https://github.com/user-attachments/assets/0f8e2966-854d-4c9d-be49-2652b581796e

To show this, I made a PoC but you may want to do this differently.
As far as I understood, there are two "issues"
1. the `SPINDLE_SPEED` entity property seems to be an Integer `{1..100}` while the `UnitSpinner` outputs a Double
    - my workaround was to add code in `setEntitySetting` to round this Double to an Integer
2. the `PERCENT` type in the `TextFieldUnit` is additionally muliplied and divided by 100
    - my workaround was to add a `PERCENT1` unit which circuments these code pathes

I haven't tested this yet on my cnc but what do you think - would it be possible to include such text field in an upcoming release?


breiler:
I made some changes here: https://github.com/mchristianl/Universal-G-Code-Sender/pull/1

PR #2900, opened 2025-09-22
Add Surfacer Plugin for Surfacing Path Generation

dimichelec:
This plugin generates rectangular surfacing toolpaths with adjustable parameters. It generates raster or spiraling paths following climb or conventional cutting directions. Surfacer opens its own tab in the plugins window of the UI for control.


breiler:
Sorry for the delay, and thanks for the contribution

dimichelec:
No problem. My pleasure.

PR #2923, opened 2025-11-05
Refactor/design settings

giro-dev:
## Refactor: Modularization of  Design Settings Panel

I've tried to encapsulate all settings depending on its "entity level" 
AbstractEntity -> transformation
Cutable -> Cuting options
Text -> Text, font

This modularization was promoted to models and panels.

### Overview
Refactored the design settings panel into modular components for improved maintainability and extensibility.

### Changes
- Split monolithic settings panel into focused, reusable components
- To handle this separation I've moved Anchor and Lock settings at AbstractEntity level.
- Improved separation of concerns for settings management
- Added changes in settings panel as undoableAction

### Testing
- All existing tests pass
- Added/updated unit tests for modularized components

### Migration Notes
- No breaking changes to public API
- Settings behavior remains unchanged for end users
- Minor changes in the User interface


ugs-platform/ugs-platform-plugin-designer/src/main/java/com/willwinder/ugs/nbp/designer/entities/Entity.java
```diff
@@ -286,4 +285,22 @@ public interface Entity {
...
+        Size currentSize = getSize();
+        setSize(new Size(width, currentSize.getHeight()));
+    }
+
+    default void setHeight(double height) {
+        Size currentSize = getSize();
+        setSize(new Size(currentSize.getWidth(), height));
+    }
+
+    void setAnchor(Anchor anchor);
```
breiler: Having the anchor and lock ratio felt a bit weird to see here. 

But I guess this is needed now in order for the transformation actions to have all data to be able to undo/redo?

giro-dev:
> I like where this is going. There are tests that are failing which needs to be looked at. I haven't had the time to look through all changes, but overall nice work!

There is more work to do... but i want to share the progress to see if It could be a good approach. So I'll work on it next weeks.

ugs-platform/ugs-platform-plugin-designer/src/main/java/com/willwinder/ugs/nbp/designer/entities/Entity.java
giro-dev: Indeed, I faced a weird behavior when undo/redo action and the anchor is not center, maybe a better approach will be save a "pivot point" instead of saving the anchor itself.

breiler:
Hey @giro-dev, have you pushed the latest changes?

I would like to add some new settings to the designer and would like to start using the changes from the PR. If you want I can try and finish it.

giro-dev:
> Hey @giro-dev, have you pushed the latest changes?

Not really because I focused on the duplicated undoables but I get stucked there...

It would be great if you could go forward with the branch. I'm having a complicated month and not sure when I could continue.

ugs-platform/ugs-platform-plugin-designer/src/main/java/com/willwinder/ugs/nbp/designer/actions/RedoAction.java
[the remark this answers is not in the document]
breiler: I think that I figured out the problem. With your group actions it is using the SelectionManagers group for applying the settings. That group is not immutable which means that after deselecting the undo/redo does not have any entities to apply the changes to.

I tried making a copy of the SelectionManager group. But that would prevent any EntityEvent to propagate up to the SelectionManager, making its visualization not up to date.

It was a good idea with the group actions, but I had a hard time making them work, so I am instead using the original single actions (RotateAction, MoveAction, ResizeAction).

breiler:
I have tested as many scenarios I could come up with and I think that it works as before. Thanks for your contribution!

PR #2933, opened 2025-11-17
Initial Implementaiton of FluidNC Setup Wizard

dnikodem84:
- Added 2 new capabilities so setup wizard does not have to show things which fluidNC does not support.
- Updated fluidNC settings to actually write settings to config.yaml at end of the process.
- Added extra endstops to the Setup Wizard ( Useful for dual_endstop configurations )

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCSettings.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCSettings.java
@@ -168,10 +168,10 @@ public class FluidNCSettings implements IFirmwareSettings {
                settingEquals("axes/z/motor1/hard_limits",true);
     }
     
-    private boolean checkMotorHasEndstop( Axis aAxis, Motor aMotor) {
-        return settingNotEqualsIgnoreCase("axes/"+aAxis.name().toLowerCase()+"/"+aMotor+"/limit_neg_pin", NO_PIN) ||
-               settingNotEqualsIgnoreCase("axes/"+aAxis.name().toLowerCase()+"/"+aMotor+"/limit_pos_pin", NO_PIN) ||
-               settingNotEqualsIgnoreCase("axes/"+aAxis.name().toLowerCase()+"/"+aMotor+"/limit_all_pin", NO_PIN);
+    private boolean checkMotorHasEndstop( Axis axis, Motor motor) {
+        return settingNotEqualsIgnoreCase("axes/"+axis.name().toLowerCase()+"/"+motor+"/limit_neg_pin", NO_PIN) ||
+               settingNotEqualsIgnoreCase("axes/"+axis.name().toLowerCase()+"/"+motor+"/limit_pos_pin", NO_PIN) ||
+               settingNotEqualsIgnoreCase("axes/"+axis.name().toLowerCase()+"/"+motor+"/limit_all_pin", NO_PIN);
     }  
     
     @Override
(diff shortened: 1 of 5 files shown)


ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCSettings.java
```diff
@@ -145,12 +158,68 @@ public List<FirmwareSetting> getAllSettings() {
...
+    @Override
+    public boolean hasZ0() {
+        return checkMotorHasEndstop("z","motor0");
+    }
+    
+    @Override
+    public boolean hasZ1() {
+        return checkMotorHasEndstop("z","motor1");
+    }    
+    private void setHardLimitEnabled(String aAxis,String aMotor,boolean enabled) throws FirmwareSettingsException{
```
breiler: Even if this is private I'd prefer to use the Axis enum instead of magic strings.

ugs-core/src/resources/MessagesBundle_cs_CZ.properties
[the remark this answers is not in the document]
dnikodem84: Reverted in my latest commit

dnikodem84:
@breiler , I have applied all review comments. It is unclear why I had a Testcase failure but committing a whitespace change made the TR pass so I am assuming that the test is intermittent.

breiler:
I have tried it out and there are a couple of things that I feel is confusing. I have just limit switches on one motor but it will also show X0, Y0 and Z0. 

I guess that the invert limit switch is a hardware configuration and can not be changed using the $CD command, so that checkbox should maybe be removed/hidden?

image

ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCSettings.java
```diff
@@ -145,24 +160,114 @@ public List<FirmwareSetting> getAllSettings() {
...
+//       return !setting.isEmpty();
+//    }
+    
+    private void setHardLimitEnabled(Axis aAxis,Motor aMotor,boolean enabled) throws FirmwareSettingsException{
+        if (checkMotorHasEndstop(aAxis, aMotor)) {
+            setValue("axes/"+aAxis.name().toLowerCase()+"/"+aMotor+"/hard_limits", enabled ? "true" : "false");
+        }
+    }
+    
+    public LinkedList<String> getUsedGpio() {
```
breiler: I prefer to use base classes such as "List" in returns. In this case it would maybe make sense to use a Set to avoid having duplicates entries.

dnikodem84:
I see this commit as a stepping stone for future improvements to the Setup Wizard for FluidNC and an alternative to the webUI.

>> I have tried it out and there are a couple of things that I feel is confusing. I have just limit switches on one motor but it will also show X0, Y0 and Z0.

My thinking was X0 was XAxis.Motor0 Endstop and X was "X axis Any Endstop", but you are right that it will add confusion in the situation where a users are unaware of "dual_endstop" configurations used for machine squaring. 

I have updated the code to show XYZ if you have one endstop per axis (No matter if one of them was on the second motor), and then IF you have at least one axis with 2 motors/1 endstops it with display the paired X0/X1 outputs (and use a spacer panel to hide the non-configured endstops) so in your scenario you will see no difference in that frame ( apart from the checkbox ).. Users who have multiple endstops need to see which one is triggered to make sure their machine is properly wired prior to homing. 

If you have a Standalone controller / ESP32 flashed with FluidNC the you can use the following config ( https://github.com/V1EngineeringInc/FluidNC_Configs/blob/main/LowRider%20CNC/Jackpot/JP1_LR/config.yaml ) to test a machine with 1 X motor but 2 motors on Y and Z. If you ONLY have a FluidNC which is configured with 

>> I guess that the invert limit switch is a hardware configuration and can not be changed using the $CD command, so that checkbox should maybe be removed/hidden?

You may have noticed that this branch adds code for tracking the used and supported pins by the current active config, each axis can be individually inverted switching /axes/x/motor0/[neg|pos|all]_limit_pin=gpio.25:low / gpio.25:high.
So for now I will add a new Capability (PER_AXIS_ENDSTOP_INVERSION) to remove the checkbox for now but a future development should add the ability for a user to define both the pin and the direction for a given endstop

ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCSettings.java
dnikodem84: Fixed in latest commit.

breiler:
Nice, thanks for the contribution.

I think that the next step should be to get the loaded config file from `$Config/Filename` and not assume it is `config.yaml`

Issue #2921, opened 2025-11-05
Improve FluidNC Machine Setup wizard


dnikodem84:
### Description

FluidNC based machines have the ability for some very advanced configuration, 
When using fluidBT then one has to either  plug their controller into a computer ( not always practical. ) or methods from the 90s ( Terminal with XModem capability ) to update machine settings perminently.

This level of configurability means  that there should be no major roadblocks in making the "UGS Machine Setup Wizard" for fluidNC machines be not only fully functional but also be able to update config.yaml to the point where the user can "square their machine" or change the homing direction without too much effort. 

for example: 

$/axes/x/steps_per_mm
$/axes/x/steps_per_mm=800

can be used to read or set the steps per MM and calibrate a machine. 

$/axes/x/homing/positive_direction

can be used to set the homing direction, 

$/axes/x/motor0/pulloff_mm=2.000
$/axes/x/motor1/pulloff_mm=0.275

can be used to provide "squaring" features to a machine which was not built completely square. 

$/axes/x/soft_limits=true
Can be used to implement soft limits. 

$/axes/x/motor0/hard_limits: true
Can be used to toggle hard limits.

PR #2938, opened 2025-11-21
Added ability to Upload / Download config.yaml to setup wizard panel

dnikodem84:
- Added new Setup Wizard panel and hid existing panel which would add confusion
- Fixed issues connecting to FluidNC controllers which had hard limits enabled ( https://github.com/bdring/FluidNC/issues/1411 )
- Fixed panel misalignment when Dual_Endstops are enabled in "limit switches" tab.
- Fixed issue with homing panel where colors were not properly reset after applying pull off amount.

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCController.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCController.java
@@ -226,7 +226,20 @@ public class FluidNCController implements IController, ICommunicatorListener {
             sendCommandImmediately(new FluidNCCommand(GrblUtils.GRBL_VIEW_PARSER_STATE_COMMAND));
         }
     }
-
+    
+    @Override
+    public void issueHardReset() throws Exception {
+        messageService.dispatchMessage(MessageType.INFO, "*** Resetting controller\n");
+        isInitialized = false;
+        positionPollTimer.stop();
+        setControllerState(ControllerState.CONNECTING);
+        resetBuffers();
+        communicator.cancelSend();
+        SystemCommand hardReset = new SystemCommand("$System/Control=RESTART");
+        sendAndWaitForCompletion(this, hardReset);
+        resetBuffers();        
+    }
+    
     @Override
     public void issueSoftReset() throws Exception {
         messageService.dispatchMessage(MessageType.INFO, "*** Resetting controller\n");
@@ -533,31 +546,6 @@ public class FluidNCController implements IController, ICommunicatorListener {
         return ControllerUtils.getCommunicatorState(state, this, communicator);
     }
     
-    private boolean detectAndUnlockHardLimits() throws InterruptedException, Exception {
-        boolean result = false;
-        if (relockHardLimits == null) {  
-            Thread.sleep(2000); 
-            issueSoftReset();
-            DetectHardLimitCommand hardLimitCheck = new DetectHardLimitCommand(false);
-            sendAndWaitForCompletion(this, hardLimitCheck);
-            
-            if (!hardLimitCheck.hasHardLimit()) {
-                hardLimitCheck = new DetectHardLimitCommand(true);
-                sendAndWaitForCompletion(this, hardLimitCheck);    
-            }
-            
-            if (hardLimitCheck.hasHardLimit()) {
-                relockHardLimits = hardLimitCheck.getRelockCommands();
-                for (SystemCommand cmd: hardLimitCheck.getUnlockCommands()) {
-                    sendAndWaitForCompletion(this, cmd);                            
-                }
-                result = true;
-            }
-        }
-        return result;
-    }
-    
-    private Set<SystemCommand> relockHardLimits;
     
     private void initializeController() {
         positionPollTimer.stop();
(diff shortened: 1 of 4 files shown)


ugs-platform/ugs-platform-plugin-setup-wizard/src/main/java/com/willwinder/ugs/nbp/setupwizard/panels/WizardPanelHardLimits.java
```diff
@@ -153,12 +153,16 @@ private void initComponents() {
         labelLimitZ1 = createLimitLabel("Z1");
         labelLimitZ1.setVisible(false);
         
-        pnlX0Spacer=new JPanel();   
+        pnlX0Spacer=new JPanel();  
+//        pnlX0Spacer.setBackground(new Color(255,0,0));
```
breiler: Remove commented out code

ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/FluidNCController.java
```diff
@@ -127,6 +129,12 @@ public FluidNCController(ICommunicator communicator) {
         this.fileService = new FluidNCFileService(this, positionPollTimer);
         this.commandCreator = new FluidNCCommandCreator();
         this.overrideManager = new GrblOverrideManager(this, communicator, messageService);
+        messageService.addListener(new MessageListener() {
+            @Override
+            public void onMessage(MessageType messageType, String message) {
+                System.err.println(message);
```
breiler: Do not log with System.out or System.err, instead use a logger.

dnikodem84: removed in latest commit

breiler:
I want to test this before merging and will not have time until a couple of hours.

I really dislike UGS to have to handle https://github.com/bdring/FluidNC/issues/1411 for multiple reasons. Ie, if the message output changes this will stop working. I also see a risk with changing the controller configuration without the user knowing (even if it is changed back and is not persisted), something that was a big problem with Easel back in the days. I am thinking of all the senders that also needs to implement this workaround in order to properly integrate with FluidNC. 

I could consider it as a short time solution but I have a feeling that this will eventually come and bite us in the back...

dnikodem84:
Please test this branch extensively.  This is a stepping stone commit before I request the closure of: https://github.com/winder/Universal-G-Code-Sender/issues/2921

I am almost finished building my newest machine and am wanting to get UGS to the point where I can take the hardware and config.yaml for a V1E LowRider ( https://docs.v1e.com/lowrider/ ) and allow UGS to configure the custom parts for my machine ( Min/Max/All endstops on various axis, Z-Probe, PWM Spindle control, maybe expose the ModBus spindle controls. )

with FluidBT I was happy using the webUI until a config.yaml issue made wifi stop working, making the BT variant more interesting, but configuration is cumbersome with the available tools so I opted to build my own.

I have standalone hardware to test with although I am working with serial so I may need to triple the timeouts make the setup wizard / updated controller work properly with the BT variant of the firmware.

I also dislike having to temporarily change the configuration because if the connection "drops"  before we can reset the setting then its possible for a user to assume everything is working properly while they have hardware endstops disabled, enter the "Web Installer" and manually enter  a "$CD=..." to overwrite their existing config ( or reconnect with UGS, go through the setup wizard, "Missing" that their endstops are disabled, and press finish ) 

I was thinking about putting up some user facing dialog (but could not find any good examples of a driver producing a warning outside of messageService.dispatchMessage ) since It feels rude to bring a swing dependancy into the driver / controller code. 

In reality this scenario is an exception caused by potentially minor issues, mostly caused during machine setup : 

1) Machine Misconfiguration ( i.e. Endstop is configured with a NO switch but user has a NC switch )

2) Machine is legitimately resting against an endstop ( in which case the user will have to go into the setup wizard, disable endstops, move the machine and then re-enable endstops _OR_ perform a homing cycle  ) 

3) ESP32 is expected to be seated in board which provides PU/PD resistors as specified by the current config but the board is disconnected from the ESP32 for "maintenance" 

4) the board is fried (ESP32 or controller.)

In every one of those scenarios it is better to let the user connect ( and potentially use the future features of the UGS Setup Wizard re-configure those pins.)

My Next branch/ Planned Pull Request brings in a "Probe Setup" page to the setup wizard:
image
(Note: Still need to implement PU/PD & LOW/HIGH buttons, as well as a clickable link to the documentation and an indicator showing if the probe is triggered so users can test the settings they are editing. )

With my (so-far incomplete) testing tonight it appears that these settings might need me to :
 
1) persist the current config to a file ( config-tmp.yaml if the controller has space ?)
2) download the config ( maybe come up with a mechansim to take the result of $CD and just save it) 
3) make the relevent changes to the .yaml. 
4) re-upload the config

If I have to "go to that length" then that Pull Request will most likely involve adding some text "[UGS: ${REST_OF_FILE_CHECKSUM}]" to a config field (like meta or name) and creating a backup of the original config if it was not created by UGS / fails the checksum.  ( which would mean the new config.yaml frame lets the user pick their "base configuration" for the rest of the process ).

dnikodem84:
Followup branch is almost ready and relies on the changes from this one, if you wish to take a look:

https://github.com/dnikodem84/Universal-G-Code-Sender/tree/ProbePanelSetup 

Pin settings do not get "accepted" by $/xxx=yyy so the followup branch contains code to manage the users configuration in a recoverable manner. 

Endstop pin configuration will be next, Followed by setting up spindle/laser/pwm control

breiler:
Thanks for your contribution

daxliniere:
"Added ability to Upload / Download config.yaml to setup wizard panel"
This is amazing, @dnikodem84, a _true_ game-changer. 🙏

Issue #2942, opened 2025-11-26
Parse 8E1 as string scalar (not number 80.0)


asomov:
### Description

It appears that SnakeYAML is parsing the specific strings “8e1” and “7e1” as floating point values (and therefore evaluating to 80.0 and 70.0).
[The ticket is created in SnakeYAML](https://bitbucket.org/snakeyaml/snakeyaml/issues/1109/cannot-use-snakeyaml-to-parse-serial)

PR #2943, opened 2025-11-26
Issue 2942: parse 8E1 as string scalar and migrate to SnakeYAML Engine

asomov:
1. Migrate to SnakeYAML Engine
2. Parse 8E1 as string scalar

To fix [issue 2942](https://github.com/winder/Universal-G-Code-Sender/issues/2942)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
     <ugs.commons-cli.version>1.4</ugs.commons-cli.version>
     <ugs.easymock.version>5.2.0</ugs.easymock.version>
     <ugs.maven-assembly-plugin.version>2.5.3</ugs.maven-assembly-plugin.version>
-    <ugs.snakeyaml.version>2.2</ugs.snakeyaml.version>
+    <ugs.snakeyaml.version>3.0.1</ugs.snakeyaml.version>
     <ugs.nashorn-core.version>15.4</ugs.nashorn-core.version>
     <ugs.jackson.version>2.15.3</ugs.jackson.version>
 
--- a/ugs-core/pom.xml
+++ b/ugs-core/pom.xml
@@ -40,8 +40,8 @@
             <version>${gson.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.yaml</groupId>
-            <artifactId>snakeyaml</artifactId>
+            <groupId>org.snakeyaml</groupId>
+            <artifactId>snakeyaml-engine</artifactId>
             <version>${ugs.snakeyaml.version}</version>
         </dependency>
         <dependency>
--- a/ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/commands/GetFirmwareSettingsCommand.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/firmware/fluidnc/commands/GetFirmwareSettingsCommand.java
@@ -19,8 +19,14 @@
 package com.willwinder.universalgcodesender.firmware.fluidnc.commands;
 
 import com.willwinder.universalgcodesender.types.CommandException;
-import org.yaml.snakeyaml.Yaml;
-import org.yaml.snakeyaml.error.YAMLException;
+import org.snakeyaml.engine.v2.api.Load;
+import org.snakeyaml.engine.v2.api.LoadSettings;
+import org.snakeyaml.engine.v2.exceptions.YamlEngineException;
+import org.snakeyaml.engine.v2.nodes.Tag;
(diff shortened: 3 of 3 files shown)


asomov:
@dnikodem84 the issue should be fixed

dnikodem84 (approved):
Looks good to me, @breiler will have to merge the PR.

asomov:
@dnikodem84 I could not deliver any unit tests - if add the unit test to cover the deviation it would make the change complete

dnikodem84:
@asomov It is not a matter of unit tests, 

The process just needs approval for new contributors and I do not have the appropriate permissions as I have only been contributing to this project for a few months.

breiler:
I added a unit test and it showed that this did not work as intended. It looks like the default FLOAT resolver needs to be removed (or add a resolver that matches 8E1 as a string):
https://github.com/snakeyaml/Universal-G-Code-Sender/pull/1/commits/2614b59d66ab42f7b931627fb06388abe266612d

I'll close this PR and recreate it with my small change.

asomov:
@breiler I wonder why not to apply your test to this PR. I could have fixed it

breiler:
@asomov sorry, I tried but I was not allowed to push to this branch. Wasn't sure if you wanted to muck around with this more than you had to.

I discovered another problem though. Netbeans uses their own class loading and it seems to not have access to the new classes and will just fail silently here:

image

In order for you to reproduce this problem you need to run the software and connect to a FluidNC device.

asomov:
@breiler I do not know how to run the software, but the build works:
`./mvnw clean test`
Can it happend that Netbeans does not use the [info file?](https://bitbucket.org/snakeyaml/snakeyaml-engine/src/master/src/main/java/module-info.java)
Can you please help me to run the software and connect to a FluidNC device?

breiler:
As far as I know it does not use the module-info, it has their own manifest files which is generated from a build task. I think their intentions are good with this system to provide modules to the platform. But it is an absolute pain to work with...

I managed to reproduce the error by adding it to the `ServiceStarter.java` so it doesn't need to connect to the device:

```
@OnStart
public class ServiceStarter implements Runnable {
    private static final Logger logger = Logger.getLogger(ServiceStarter.class.getName());

    @Override
    public void run() {
        GetFirmwareSettingsCommand getFirmwareSettingsCommand = new GetFirmwareSettingsCommand();
        getFirmwareSettingsCommand.getSettings();
    ...
```

Which produced this error:
```
SEVERE [org.openide.util.RequestProcessor]: Error in RequestProcessor com.willwinder.ugs.nbp.core.lifecycle.ServiceStarter
java.lang.ClassNotFoundException: org.snakeyaml.engine.v2.exceptions.YamlEngineException
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:525)
Caused: java.lang.NoClassDefFoundError: org/snakeyaml/engine/v2/exceptions/YamlEngineException
	at com.willwinder.ugs.nbp.core.lifecycle.ServiceStarter.run(ServiceStarter.java:51)
	at org.openide.util.RequestProcessor$Task.run(RequestProcessor.java:1403)
	at org.netbeans.modules.openide.util.GlobalLookup.execute(GlobalLookup.java:45)
	at org.openide.util.lookup.Lookups.executeWith(Lookups.java:287)
[catch] at org.openide.util.RequestProcessor$Processor.run(RequestProcessor.java:2018)
```

breiler:
There is a run profile "UGS-Platform" available if you use IntelliJ. Or you can run the `./run_platform.sh`. 

But you don't have to look at this anymore, I can see if I manage to add the new library to the classpath.

asomov:
I did not quite catch how it help me to reproduce the iisue. Anyway, if there is another way to enclude/exclude classes we have to instruct UGS-Platform how to find and use the classes.

asomov:
It requires Maven to be installed. I will deliver a fix to use the Maven wrapper:
```
 Universal-G-Code-Sender % ./run_platform.sh
./run_platform.sh: line 2: mvn: command not found
```

breiler:
Gah, I have spent all morning debugging this. 

I tested building a package and running it outside the IDE and it worked without any problems.

So I then tried making sense of why it didn't work in the IDE, reloading the maven project, looking through all kind of settings. Then I finally cleared the IntelliJ:s caches, recompiled and now it works... 🤷🏻‍♂️ 

Sorry for the noise...

Issue #2959, opened 2025-12-15
Add ability to print designs from the designer


dnikodem84:
### Description

When designing a part or enclosure it is sometimes useful to be able to print out the part on paper prior to cutting it to make sure it lines up with other things . 

As of right now it is not possible to use print directly from UGS and the PNG exporter does not correctly render the exported PNG files ( due to the affine transform being used being pretty incorrect )

PR #2960, opened 2025-12-15
- Added ability to print from UGS Designer.

dnikodem84:
1) Added new PrintDesignAction with all appropriate plumbing and basic pagination and correct scale. 
2) Fixed issues in Drawing.GetImage causing the png to be incorrect in most circumstances.


dnikodem84:
this should fix: https://github.com/winder/Universal-G-Code-Sender/issues/2959

PR #2965, opened 2025-12-24
Change base URI in PendantUI to allow external access

ldab:


breiler:
What platform are you using where it is not accessible publicly? 

Jetty will, without extra configuration, automatically bind to all interfaces. So your changes should have no effect, but will be more clearer on the intention of the code so I will merge it.

ldab:
> What platform are you using where it is not accessible publicly?
> 
> Jetty will, without extra configuration, automatically bind to all interfaces. So your changes should have no effect, but will be more clearer on the intention of the code so I will merge it.

I’m running on Debian 12 with Tailscale, Cisco VPN, and some containers… for some reason using 0.0.0.0 makes it more reliable, before it’s kind of hit and miss

PR #2967, opened 2025-12-28
- Added rename action to main popup menu for tree

dnikodem84:


breiler:
The rename does not work if the action is initialized without the entity tree. For instance if you select a shape and press F2.

I suggest that we decouple the rename with a EntitiesTreeController which is registered to the central lookup.

https://github.com/dnikodem84/Universal-G-Code-Sender/pull/2

PR #3004, opened 2026-01-23
Add support for Excellon drill file

breiler:
It will group all hole sizes to its specific group:
image


atrabit:
What about ovals? In the screenshot above the parser shows two holes while there is an oval. So it will require more attention from the user to check if the ovals will be done with drill or a end-mill bit and at what depth.

breiler:
Thanks, I think I managed to separated the routing operations in #3004

Issue #1842, opened 2022-03-21
Visualizer: Assign middle-button drag to pan window


daxliniere:
### Description

It would also be cool if we could use middle-button drag to pan the Vis window, instead of (as well as) SHIFT+left button. Many other programs use this method.

PR #3018, opened 2026-02-02
Visualizer: Assign middle-button drag to pan window #1842

atrabit:
Resolving #1842 request: adding pannig in Visualizer by middle or right mouse button and Reset view when lost in zooming by middle buton click.


atrabit:
Increasing arc render quality for small arcs eg. thread milling:
ARC_SEGMENT_LENGTH=0.8
Screenshot 2026-02-02 233010
ARC_SEGMENT_LENGHT=0.1
Screenshot 2026-02-02 233411

breiler:
Sorry for the delay. I was worried that this change would cause the memory to explode on larger arcs.

I tested with this and could hardly notice any Heap memory difference. We'll go with this until it can either be configured through a setting or we change the algorithms to split arcs to fixes line segments.
```
G17 G20 G90 G94 G54 M0 M5 M9
G0 Z250 F5
X-50 Y0.
G1 Z0.
(arcs...)
G2 X0. Y-50 I50 J0.
G2 X-50 Y0. I0. J50
```

atrabit:
I'll try to figure out how to make it dynamiclally with the actual zoom level. It is all about how it looks on the screen eventually,

PR #3057, opened 2026-03-20
Raster: Power Curve + Performance Improvements

giro-dev:
### Summary
Adds a **power curve editor** to the Raster image settings, allowing users to compensate for the non-linear relationship between pixel brightness and actual laser power output. Also includes significant raster processing performance improvements.

---

### New feature — Power Curve

Laser engravers do not burn linearly with power input: a 50% power value does not produce 50% burn intensity. The power curve lets users define a custom brightness → power mapping to compensate for this.

The curve editor (`CurveCanvas`) works similarly to Photoshop's curves tool:
- Click to **add** a control point
- Drag to **move** a control point
- Ctrl+click or double-click to **remove** a control point (endpoints are fixed)
- Cursor changes to `+`, `−` or move symbol to hint the current action
- Tooltip appears after 2 s of inactivity with usage instructions

The curve is a **Catmull-Rom spline** rendered via Java2D `Path2D`, and is sampled into a 256-entry LUT applied to every pixel during raster processing. The visual curve and the LUT share the same path builder (`MonotoneCubicSpline`), so what the user sees matches what the laser does.

image
image

---

### Raster processing performance

The per-pixel pipeline (brightness → contrast → gamma → invert → levels → power curve) was refactored to run **once** at setup time instead of per pixel:

1. **Master LUT** — all adjustments pre-computed into a single `int[256]` lookup. The hot loop becomes one luminance calculation + one array lookup per pixel instead of 6+ floating-point operations.
2. **Bulk pixel I/O** — replaced per-pixel `getRGB(x,y)` / `setRGB(x,y)` with `getRGB(0,0,w,h,...)` bulk reads and direct `DataBuffer` writes, eliminating ~2M method calls on a 1024×1024 image.
3. **Ink mask via DataBuffer** — `getOrCreateProcessedInkMask` reads/writes through `DataBufferByte`/`DataBufferInt` directly.

---

PD: I was not able to test the real output in a laser machine because mine is not operative. bur since the preview is really the final output i think the result has to be fine


breiler:
Really nice, thanks for the contribution! I can give it a test, but it will have to wait until tomorrow or Sunday.

breiler:
I'd call that a success. These are tiny 30x40mm images on birch plywood. The image to the left is the orignal image without any adjustments. The middle I tried just increasing the contrast. The image to the right, I increased the contrast slightly and applied a curve that I believed would do a good job (just winging it):

PXL_20260322_150602034 MACRO_FOCUS

Screenshot from 2026-03-22 16-12-16

If I would make a test pattern I would probably find a good curve that would work well with my laser in most cases.

giro-dev:
> I increased the contrast slightly and applied a curve that I believed would do a good job (just winging it):
> 
> PXL_20260322_150602034 MACRO_FOCUS
> 
> Screenshot from 2026-03-22 16-12-16
> 

It looks good!
Thanks for the real testing!!

PR #3089, opened 2026-04-22
Fixed crash on macOS when opening designer

dnikodem84:
Fixed crash where SvgIconLoader.class.getResourceAsStream("/img/zoom-in.svg") was doing a getResourceAsStream against ugs-core's classloader, but the icons actually live in ugs-designer (which sits in a downstream module classloader ugs-core can't see under the NetBeans Platform), so loadImageIcon came back empty, .orElse(null) handed a null to Toolkit.createCustomCursor, and mac's CCustomCursor NPE'd on File -> New Design


breiler:
Thanks

PR #3090, opened 2026-04-22
Fixed issue where selection did not update after clicking on shape

dnikodem84:
Selection is only populated itself from SelectionListener events. But DesignerTopComponent.onSelectionEvent opens the settings pane via SwingUtilities.invokeLater, so SettingsTopComponent.componentOpened runs after the SelectionEvent has already been dispatched. A fresh SelectionSettingsPanel then  registers as a listener with no way to learn about the just-selected shape, leaving the pane empty and disabled


dnikodem84:
There appear to be issues with the "Shape settings" after i refreshed from main today. If you have any idea what the remaining issue is I can attempt to fix next week otherwise I might have to revert to an older version. 

image

dnikodem84:
Todays commit has my fork properly loading the shape settings in the designer when launching from ./run_platform.sh 

image

PR #3091, opened 2026-04-22
Forced settingsChanged to run in the EDT preventing threading issues

dnikodem84:
Before 60a684c1, the handler held a Future<?> settingsChangedFuture guard so only one rebuild was ever scheduled at a time; subsequent events while a rebuild was pending were dropped, and the guard was cleared at the end of settingsChanged().

 The synchronized(settingsLock) block is a no-op (it guards nothing that is actually shared). Every SettingChangedEvent now schedules a new rebuild task.

 The user's keyboard shortcuts fire actions that produce SettingChangedEvents in rapid succession. After the 2-second delay, N queued rebuilds all run concurrently on the 10-thread pool, each calling removeAll() and then add(...) on the same buttonGridPanel. Interleaved execution on two+ worker threads races on Swing's (non-thread-safe) container list


breiler:
Excellent, thanks!

PR #3092, opened 2026-04-22
Update Printer Functionality

dnikodem84:
The ability to print a design out before carving saves a lot of material / time as there is no replacement for a physical version of a design in terms of validation and traceability. (I.e. it is hard to argue with a signed and timestamped pre-production printout )

The "Print" feature is a lifesaver but it could be a little bit more professional so this change-set adds some configuration options to the printout as well as adding metadata which most users who would use "printing" as part of a production/design pipeline would fine useful in their physical copies (Timestamp, Author, Scale Factor, etc. )

There is future development to do here (I.e. bifuricate printing and print-preview, add axis to the border, engineering templates only work with single page.)


dnikodem84:
image

This is a solid first pass at a print preview.

breiler:
Nice!

PR #3093, opened 2026-04-24
Added Tool Library

dnikodem84:
- "Pick from Library..." button added to the existing Tool Settings dialog. Picking a tool populates the fields and shows its name; editing a field away from the library value  
  turns it amber until it matches again. The picker also has a "Manage Library..." button for adding/editing tools.
  - Tool Library manager — list on the left, editor on the right, with Add / Duplicate / Delete / Revert. Built-in tools can be edited and reverted, but not deleted.              
  - 41 pre-seeded tools: Upcut, Downcut, Ball, Compression, Straight in 1/16", 1/8", 1/4", 1/2", 1" sizes; V-bits in 30°, 60°, 90° across the same sizes; plus a "Custom" entry.   
  - Per-tool diameter units stick with the tool — 1/4" endmills always display as 0.25 in, 3 mm endmills always as 3 mm. Toggling the unit in the editor converts the value.       
  - Feed / plunge / depth stay in machine units and now show unit labels (mm/min, mm) next to each field.                                                                          
  - Library persists across projects at ~/.ugs/tool-library.json.                                                                                                                  
  - Project files remember which tool they were built with. On opening a project whose tool differs from your library, UGS prompts: use the project's tool for this session, update
   the library, or keep your library version. If the project's tool isn't in your library at all, it offers to import it.                                                          
  - Existing projects load unchanged — no prompt, no overwriting of your current tool settings.


dnikodem84:
This is a first step towards some significant depthmap improvements ( roughing and finishing pass, proper V-Carving with  Minkowski dilation + iso-scallop ). Those depthmap improvements will add an optional tool to each cuttable, and add support for exporting / running one tool at a time. 

At that point it should be pretty trivial to add support for "T" commands if the machine supports them.

breiler:
I gave it a try, nice work. I found a couple of issues that would like having fixed:

* The step over percent is scaled 100x wrong. If I set 30% it will become 3000% in the tool library:
image


* I would expect the text boxes to not be yellow when using the custom tool.
* The editing UI gets squashed (have only tried on linux):
image

breiler:
Sorry for the late response, it has been a busy week for me.

I gave this some testing again and it will still show the fields to have changed when I select a predefined tool and open up the dialog again.

https://github.com/user-attachments/assets/7cd83f5c-ad3e-4ed3-83e1-e5cac652b67d

PR #3117, opened 2026-05-30
Treat homing as enabled for non-zero values

### LINKED ISSUE #3114: Allow homing when `($22 & 1) != 0` to support grblHAL
### Description

As an extension, GrblHAL allows the values of `$22` (which in Grbl is a boolean "enable homing" setting) to be larger than 1. It uses it as a bitmask to store additional homing settings.

But UGC currently refuses to home if it's not set to exactly `1`, thinking that homing is disabled.

It would be nice to instead allow homing if the first bit is set (and "set homing enabled" should similarly only modify the first bit). This changes nothing for the original Grbl users, and fixes homing on GrblHAL.

I can submit a PR if needed, seems simple enough.

breiler:
Fixes #3114

### DIFF
--- a/ugs-core/src/com/willwinder/universalgcodesender/firmware/grbl/GrblFirmwareSettings.java
+++ b/ugs-core/src/com/willwinder/universalgcodesender/firmware/grbl/GrblFirmwareSettings.java
@@ -190,16 +190,13 @@ public class GrblFirmwareSettings implements IFirmwareSettings {
 
     @Override
     public boolean isInvertDirection(Axis axis) throws FirmwareSettingsException {
-        switch (axis) {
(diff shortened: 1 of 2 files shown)


HolyBlackCat:
I guess it doesn't really matter, but grblHAL refuses to home if `$22` has zero-th bit unset, even if `$22` is positive. The more accurate condition would be `($22 & 1) != 0`.

breiler:
Yeah, I am not so keen to adding more GrblHAL special logic to the original GRBL firmware profile. I will probably eventually build this https://github.com/winder/Universal-G-Code-Sender/issues/1468. Until then I think this will have to suffice.

Issue #1317, opened 2019-12-10
Supporting Tool Offsets


TheRealJAnderson:
## Feature request

I want to use a tool offset probe with my CNC with UGS as the controller and don't find that UGS supports tool offsets. Does it and I just don't know about it?

Punmageddon:
I wrote a little macro that can handle tool-length-offsets for me.
Maybe this will help:


Basically you run it when you start the cnc and then you run it when you have put in a new bit.

You need to install a fixed probe. I just use a normal button (repeatability is +-0.01mm).

This macro uses G92 offsets and the G59 work coordinate system, so you can not use them for your normal cnc work.
You need to set G59 zero points over your fixed probe button. Thats how you configure the position of the probe.

```
G92.1;
G90;
G53 G0 Z0 F3000;
G59;
G0 X0Y0;
G0 Z0;
G91;

G38.2 Z-90 F300;
G0 Z2;
G38.2 Z-10 F30;
G4 P.25;
G38.4 Z10 F20;
G4 P.25;
G38.2 Z-10 F10;
G4 P.25;
G38.4 Z10 F5;
G4 P.25;

G92 Z0;

G0 Z5 F3000;
G90;
G53 G0 Z0 F3000;
G54;
```

It just uses normal gcode so it should really run in any gcode sender.
Honestly this should maybe be a built in feature. That way you might not have to lose G92 and G59.

mirko91:
Hi! Is this feature going to be built in on UGS?

breiler:
I am closing this, UGS will now try and intercept M6 tool change commands and let the user rezero the machine and continue the job: #3149

PR #3149, opened 2026-07-24
Add support for tool changes

breiler:
Added a new mechanism for interception of commands. When the stream encounters a command it is looking for it will make sure that the program runs up to that point and settles into an IDLE state. After that a user operator stage can take place where certain operations may be made such as a tool change. Once that is done the stream may continue.

The tool change process will raise the spindle to the highest place of the gcode and stop the spindle. After that the user may jog around and change the tool, then jog and reset Z and then continue the stream. There is also an option to run the probe.

https://github.com/user-attachments/assets/14f15840-ab09-4777-ab12-1edce8712b84

The tool change operation is turned on by default, if your controller can handle tool changes, the interceptor may be turned off with the "Handle tool changes" setting:
Screenshot from 2026-07-24 17-08-15


mirko91:
Hello, so inside that prompt the macro from #1317 has to be run in order to have the tool offset?

breiler:
As I have built it, UGS should take care of the tool offsets. No need to define macros or scripting.

Run the job with a M6 tool change and UGS will ask you to change the tool and reset the new zero (or you can use probing if you have that set up) and just continue the gcode program.

ThorbenH:
It's been a while since I wrote that macro.
Basically you equip your machine with a Tool-Length-Sensor (a button also works) -> Mounted to the base of the work area.
Then you set your G59 origin above that sensor.
Then you use the macro with your first tool of the job.

And then on any toolchange you can use that macro again.

It uses G92 work offsets. That way you dont have to manually do the work of measuring and rezeroing.

I have not tested the macro in combination with the new UGS feature and am also unable to do that right now as I dont currently have a functional CNC.

You should never run a macro you dont fully understand. Just google the GCodes one by one as you read them. Then once you understand the macro feel free to test it and tell us the results.

PR #3155, opened 2026-07-27
Add a Vcarve tool path

breiler:
Added a new toolpath that will use the vbit angle setting to determine how deep it should move the tool to produce the desired line width. If the tool can't fill the width it will do a pocket instead.

image

image


atrabit:
Thanks, that's a very welcome improvment!

PR #3178, opened 2026-08-19
Backlash compensation

breiler:
I built the https://github.com/openBrushograph/openBrushograph_hardware/ and got issues with backlash from the 28BYJ-48 stepper motors.

So I tried implementing a backlash compensator which will apply a counter backlash movement when the motors moves in another direction.

Settings are available in in the controller options, enable "Backlash compensation" and set the size of the backlash in millimeters:
image

This is the test pattern I tried plotting:
image

Which gave this result, size wrong, rounded rotated square corners, circle with straight sides:
image

When applying the backlash compensation the gcode looks all mangled up:
image

But gives this result, looks much better:
image

Pretty darn cool:
image


dusjagr:
Wow!
thanks alot... we always had to run the code throug an external script.

PR #3194, opened 2026-09-06
Add a dashboard-style touch UI alongside the classic web pendant

qviper:
## Summary

Adds a second front end for the embedded pendant server (`ugs-pendant`), served alongside the existing classic pendant at a new `/dashboard` path on the same Jetty server. Targets a 13-21in landscape touchscreen mounted at the machine: an always-visible layout with a big DRO, a jog pad with Z controls spaced away from the XY pad (reduces accidental touches on a touchscreen), macros, a console, a gcode editor, and a live 3D toolpath visualizer.

- Same React/TS/Vite stack as the existing pendant webapp, reusing its REST/WebSocket API and most of its service/model/store layer as a starting point.
- Gcode editor: CodeMirror 6 with a small hand-written gcode syntax-highlighting mode (no off-the-shelf CM6 gcode language exists), save-to-file, read-only while a job is running.
- 3D visualizer: Three.js, fed by a new `/api/v1/visualizer/getToolpath` endpoint that reuses `GcodeViewParse`/`LineSegment` - the same parsing path the desktop 3D view uses - instead of re-implementing gcode parsing in JavaScript.
- New `/api/v1/files/getFileContent` and `saveFileContent` endpoints for the editor, gated to files already listed by `getWorkspaceFileList` to avoid arbitrary filesystem access from a client-supplied filename.
- The classic pendant at `/` is unchanged.

## Notable bugs found and fixed while building this (both caught by testing against the real server, not just a mock)

1. The reused `services/*.ts` used relative `fetch("api/v1/...")` calls, which resolve to `/dashboard/api/v1/...` when the page is served from `/dashboard/` and 404 on the real server. Switched to absolute `/api/v1/...` paths.
2. `GcodeViewParse`'s first segment(s) can have `NaN` coordinates (the parser's undefined starting position before the first real move), which Jackson serializes as the string `"NaN"` - this would have silently corrupted the WebGL vertex buffer. Filtered out server-side in `VisualizerResource`.

## Test plan

- [x] `mvn -pl ugs-pendant -am install` builds cleanly (Java 25)
- [x] `npm run build` (webapp-dashboard) - `tsc` + `vite build` clean
- [x] Ran the real embedded server via `ugs-cli -d` (no hardware needed) and verified against it directly (not just a mock):
  - `/` (classic pendant) and `/dashboard/` (new dashboard) both served correctly from the same server
  - `/api/v1/...` reachable from both, including the new `getFileContent`/`saveFileContent`/`getToolpath` endpoints
  - Path-traversal attempt on `getFileContent` (`file=../../../etc/passwd`) correctly rejected with 404
  - Opened a real workspace file and confirmed `getToolpath` returns correct, real parsed geometry
- [x] Interactive pass against a mock backend (included at `ugs-pendant/mock-backend/`, useful for iterating on this UI without a full Java build or real hardware): jog, zero, home, macros, console send, file open/run, gcode editor load+edit+save, 3D visualizer render+orbit

## Known follow-ups (not blocking, happy to address in this PR or a follow-up)

- The narrow-screen (<900px) fallback is a scrollable single column rather than a tabbed layout.
- The dashboard's JS bundle is ~1.25MB (Three.js + CodeMirror + Bootstrap + FontAwesome) - could be code-split (lazy-load the editor/visualizer) if that matters.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


breiler:
Thanks for this!

Really cool to see a visualizer in the webapp and we should add this. I also liked a larger view that can be seen on a desktop or a pad. I did however not like the duplication. 

If you could incorporate this into the old pendant UI and make a responsive state for larger screens and show this kind of view I'd consider merging it.

breiler:
You have made changes in many parts of the software + broken the build scripts. I like some of the changes, but since this PR touches so many different things I can't accept it.

qviper:
yes, sorry. Never meant to send a PR at this point — it's not ready. It does touch quite a bit of code in pendant, and a little in CLI, but not much (if anything) in platform. Works very well on my home system though, continuing progress.

PR #3204, opened 2026-09-11
Stop laser output on rapid moves

breiler:
Some controllers do not support the GRBL $32=1 laser mode. This fix should resolve that by always turning of laser between rapids.


daxliniere:
Oof! Good catch, Joacim!

breiler:
GRBL and FluidNC handles this with laser mode enabled. This is for legacy controllers or where they have removed the support for some reason.

massimocarbognin:
That's what i was waiting for. Is it already on latest release?

